PR #137, opened 2021-03-08
chore: Ignore submodules when using clang-format

### LINKED ISSUE #59: Informat.sh script, blacklist third party code and troubleshoot `.clang-format` config usage
The `format.sh` script currently formats the files in our submodules. We should either blacklist anything in `runtime/thirdparty`. Additionally, it's unclear to me if the .clang-format file is actually being ingested. I believe that I've successfully been able to modify settings here and impact the formatting logic, but while investigating this issue, this should be confirmed and then the  `indentation` keys should be looked at a bit close

There might be a conflict here. How does `TabWidth` and `IndentWidth` interact?
IndentWidth: 8
TabWidth: 8
UseTab: ForIndentation

bushidocodes:
Resolves #59 

Please check the following:
1. That your clang-format installation is picking up our project configuration. Running `./format.sh` should not trigger changes beyond what I have added to this PR
2. That your code editor is configured to run clang-format on save, such that it does not conflict with the formatting performed by this script
3. That running `./format.sh` does not format code in submodules. Each repo should be responsible for handling this locally.

### DIFF
--- a/format.sh
+++ b/format.sh
@@ -25,6 +25,6 @@ if ((major < required_major)) || ((minor < required_minor)) || ((patch < require
   exit 1
 fi
 
-find ./runtime -type f -path "./*.[ch]" |                             # Match all *.c and *.h files in ./runtime
-  grep --invert-match -E "./runtime/jsmn/*|./runtime/http-parser/*" | # Excluding those in the jsmn or http-parser submodules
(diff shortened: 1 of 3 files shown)


emil916:
Ah, I forgot to write back about this. I tested clang-format with version 9 and the `./format.sh`  worked fine!

PR #237, opened 2021-05-27
feat: deadline description to real world workload mix

bushidocodes:


runtime/experiments/workload_mix_realworld/spec.json
```diff
@@ -0,0 +1,60 @@
+{
+	"active": true,
+	"name": "ekf",
+	"path": "ekf_wasm.so",
+	"port": 10000,
+	"expected-execution-us": 5000,
+	"relative-deadline-us": 354,
```
emil916, line 7: Shouldn't rel deadline be greater than expected exec?

emil916 (commented):
Sorry for getting back late. It looks all good to me. The only thing is the http MIME types in the spec json files seem not be all in place...

PR #238, opened 2021-06-01
Fine tune experiments

### LINKED ISSUE #230: Linux host prio settings
The current Linux scheduler code is commented out after inadvertently starving key components of the the operating system, leading to unresponsiveness. The settings should be adjusted down slightly such that the system remains responsive, but all pinned threads consume nearly all of their cores.

bushidocodes:
Resolves #230


runtime/include/listener_thread.h
```diff
@@ -5,7 +5,7 @@
 #include "generic_thread.h"
 #include "module.h"
 
-#define LISTENER_THREAD_CORE_ID 0
+#define LISTENER_THREAD_CORE_ID 1
```
emil916, line 8: What was the reason to make this change?

emil916: Oh I think I got it, to leave core 0 for the system etc.

bushidocodes: Yep, that's pretty much it!

PR #254, opened 2021-06-07
feat: Initial attempt at non-C example

bushidocodes:


gparmer:
This actually works?

bushidocodes:
@gparmer - It composed successfully with `awsm`, but it bombs out when sledge tries to load it because the `wasmf___init_libc` symbol doesn't exist (since it's not C). I mostly opened this in case Xiaosu was interested in investigating this further since she was asking about code other than C. I'd be curious to see what happens if we simply push forward without calling `wasmf___init_libc` if the symbol doesn't exist.

PR #257, opened 2021-06-11
Percentile Cleanup, Processor Speed Hack, and Experimental Config

bushidocodes:
This PR contains
- Various experimental config changes for the last run I did on Friday
- Better Documentation
- A more robust series of utility functions for generating percentile tables


runtime/experiments/payload/run.sh
```diff
@@ -140,9 +140,9 @@ process_results() {
 		awk '
 		BEGIN {
 			sum = 0
-			p50 = int('"$oks"' * 0.5)
-			p90 = int('"$oks"' * 0.9)
-			p99 = int('"$oks"' * 0.99)
+			p50 = int('"$oks"' * 0.5) + 1
```
emil916, line 143: No clue what this is for

bushidocodes: I added this offset because Awk uses 1-based indices for rows, so the first row is 1, not 0. However, this ended up getting rendered moot by the checks I have for either rendering a number or `*` if there isn't enough sample data. This code will end up replaced by calls to the bash library functions in `percentiles_table.sh` when this draft PR is done.

emil916 (approved):
Looks good. Liked the new percentile table stuff.

runtime/experiments/bash_libraries/percentiles_table.sh
```diff
@@ -0,0 +1,77 @@
...
+			# Empty pattern matches all rows
+			             { sum += $0 }
+			NR==1 		 { min = sample_size  > 0   ? sprintf(format_string, $0) : invalid_number_symbol }
+			NR==p50_idx  { p50 = sample_size >= 3   ? sprintf(format_string, $0) : invalid_number_symbol }
+			NR==p90_idx  { p90 = sample_size >= 10  ? sprintf(format_string, $0) : invalid_number_symbol }
+			NR==p99_idx  { p99 = sample_size >= 100 ? sprintf(format_string, $0) : invalid_number_symbol }
+			NR==p100_idx { p100 = sample_size > 0   ? sprintf(format_string, $0) : invalid_number_symbol }
+
+			END {
+				mean = sample_size > 0 ? sprintf(format_string, sum / NR) : invalid_number_symbol
```
gparmer, line 72: This is great. I think that the mean is the most useful to get an intuition about system utilization for a given request rate.

runtime/experiments/deadline_description/README.md
```diff
@@ -22,28 +22,35 @@ cd ../workload_mix_realworld/
...
 7. Repeat for steps 5,6 for each desired scheduling policy.
 
 8. The results are on the server. You may want to copy them to your client to more easily inspect.
 
 ```sh
-scp -r sean@192.168.7.26:~/projects/sledge-serverless-framework/runtime/experiments/workload_mix_realworld/res/myrun ./res/myrun
+scp -r sean@192.168.7.26:~/projects/sledge-serverless-framework/runtime/experiments/workload_mix_realworld/res/myrun/* ./res/myrun
 ```
+
+9. If you are assembling a spreadsheet, you might want to also copy the results from the `deadline_description` experiment. The `deadlines.csv` file and the execution_time files are useful for understanding the performance characteristics of a single workload.
```
gparmer, line 56: Thank you for keeping up with all of this documentation!

runtime/experiments/workload_mix_realworld/run.sh
```diff
@@ -105,7 +106,7 @@ run_experiments() {
 	local -i batch_id=0
 	local -i roll=0
 	local -ir total_iterations=10000
-	local -ir worker_max=30
+	local -ir worker_max=50
```
gparmer, line 109: I'm guessing these can be overridden by env variables?

runtime/experiments/concurrency/latency.gnuplot
```diff
@@ -13,7 +13,9 @@ set yrange [0:]
 
 set style histogram columnstacked
 
-plot 'latency.dat' using 1:2 title 'p50', \
-     'latency.dat' using 1:3 title 'p90', \
-     'latency.dat' using 1:4 title 'p99', \
-     'latency.dat' using 1:5 title 'p100', \
+plot 'latency.dat' using 1:8 title 'p100', \
```
gparmer, line 16: Would love to see the output of this. Would also be nice at some point in the future to have an actual CDF; not worth working on this now.

gparmer (approved):

gparmer:
I'd imagine that you want the processor speed stuff to be an env var. @emil916 can determine the plans there and make a decision.

runtime/experiments/workload_mix_realworld/run.sh
bushidocodes: Not the way it's written currently, but something like `local -ir worker_max=${WORKER_MAX:-50}` would do with trick with a default value if it's not defined. Likely a better approach than noise diffs!

runtime/experiments/concurrency/latency.gnuplot
bushidocodes: I attempted to port this to some of the other experiments since they now produce a consistent output via the utility functions, but the way this is written seems to depend on the x axis being a numeric value. Based on what's checked into this repo, I don't think I got far enough with gnuplot to figure out how to have a set of strings on the x axis. I imagine some sort of google search of "gnuplot candlestick" would be useful.

gparmer:
I think that @emil916 can add envvars as he finds useful. I think that this is good to apply.

PR #261, opened 2021-06-22
Fixed the Dynamic read of the Number of CPU Cores

### LINKED ISSUE #260: Sledgert hangs when high number of worker cores are set
Sledge runtime hangs and does not respond to a client request at all when `NWORKERS` are set to a high number. This happens on a machine that has a two-socket CPU, each socket having 10 cores. SMT is off.

How to reproduce the problem:

Server side:
`~/sledge-server/runtime/experiments/workload_mix_realworld$ SLEDGE_NWORKERS=16 ./run.sh -s -e=fifo_nopreemption.env`

Client side:
`~/sledge-client/runtime/experiments/workload_mix_realworld$ curl --data-binary "@cifar10/airplane1.bmp" "http://10.10.1.1:10000" `  
(there 10.10.1.1 is the server's IP addr.)

And this is the per core CPU utilization info. As seen, from core c2 up to c18 (total of 16 cores) are the worker cores and core c1 is the listener. Core c0 is the main thread's core.
Screenshot 2021-06-20 025901

emil916:
Closes #260

### DIFF
--- a/runtime/Makefile
+++ b/runtime/Makefile
@@ -5,9 +5,6 @@ $(error Unsupported Architecture. Supports x86_64 and aarch64, found $(ARCH))
 endif
 endif
 
-# TOTAL_CORES := $(shell getconf _NPROCESSORS_CONF)
-# PAGE_SIZE := $(shell getconf PAGESIZE)
-
 # Compiler Settings
 CC=clang
 CC_OPTIONS = -O3 -flto -g -pthread -D_GNU_SOURCE
--- a/runtime/Makefile
+++ b/runtime/Makefile
@@ -5,8 +5,8 @@ $(error Unsupported Architecture. Supports x86_64 and aarch64, found $(ARCH))
 endif
 endif
 
-TOTAL_CORES := $(shell getconf _NPROCESSORS_CONF)
-PAGE_SIZE := $(shell getconf PAGESIZE)
+# TOTAL_CORES := $(shell getconf _NPROCESSORS_CONF)
+# PAGE_SIZE := $(shell getconf PAGESIZE)
 
 # Compiler Settings
 CC=clang
@@ -81,8 +81,8 @@ BINARY_NAME=sledgert
 
 # Sets a flag equal to the processor architecture
 CFLAGS += -D${ARCH}
-CFLAGS += -DNCORES=${TOTAL_CORES}
-CFLAGS += -DPAGE_SIZE=$(PAGE_SIZE)
+# CFLAGS += -DNCORES=${TOTAL_CORES}
+# CFLAGS += -DPAGE_SIZE=$(PAGE_SIZE)
 
 # Sandboxes running on Sledge always use WebAssembly linear memory
 CFLAGS += -DUSE_MEM_VM
--- a/runtime/Makefile
+++ b/runtime/Makefile
@@ -81,8 +81,6 @@ BINARY_NAME=sledgert
 
 # Sets a flag equal to the processor architecture
 CFLAGS += -D${ARCH}
-# CFLAGS += -DNCORES=${TOTAL_CORES}
-# CFLAGS += -DPAGE_SIZE=$(PAGE_SIZE)
 
 # Sandboxes running on Sledge always use WebAssembly linear memory
 CFLAGS += -DUSE_MEM_VM
--- a/runtime/experiments/applications/ekf/one_iteration/spec.json
+++ b/runtime/experiments/applications/ekf/one_iteration/spec.json
@@ -1,6 +1,6 @@
 {
   "active": true,
-  "name": "resize",
+  "name": "ekf",
   "path": "ekf_wasm.so",
   "port": 10000,
   "expected-execution-us": 5000,
--- a/runtime/src/main.c
+++ b/runtime/src/main.c
@@ -345,8 +345,6 @@ main(int argc, char **argv)
 
 	printf("Runtime Environment:\n");
 
-	// memset(runtime_worker_threads, 0, sizeof(pthread_t) * RUNTIME_WORKER_THREAD_CORE_COUNT);
-
 	runtime_processor_speed_MHz = runtime_get_processor_speed_MHz();
 	if (unlikely(runtime_processor_speed_MHz == 0)) panic("Failed to detect processor speed\n");
 
--- a/runtime/src/main.c
+++ b/runtime/src/main.c
@@ -61,7 +61,7 @@ runtime_allocate_available_cores()
 
 	/* Find the number of processors currently online */
 	runtime_total_online_processors = sysconf(_SC_NPROCESSORS_ONLN);
-	printf("\tCore Count: %u\n", runtime_total_online_processors);
+	printf("\tCore Count (Online): %u\n", runtime_total_online_processors);
 
 	/* If more than two cores are available, leave core 0 free to run OS tasks */
 	if (runtime_total_online_processors > 2) {
@@ -259,10 +259,9 @@ log_compiletime_config()
 	printf("\tArchitecture: %s\n", "x86_64");
 #endif
 
-	int ncores = NCORES;
-	printf("\tTotal Cores: %d\n", ncores);
-	int page_size = PAGE_SIZE;
-	printf("\tPage Size: %d\n", page_size);
+#ifdef PAGE_SIZE
+	printf("\tPage Size: %lu\n", PAGE_SIZE);
+#endif
 
 #ifdef LOG_HTTP_PARSER
 	printf("\tLog HTTP Parser: Enabled\n");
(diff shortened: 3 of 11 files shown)


runtime/Makefile
```diff
@@ -5,8 +5,8 @@ $(error Unsupported Architecture. Supports x86_64 and aarch64, found $(ARCH))
 endif
 endif
 
-TOTAL_CORES := $(shell getconf _NPROCESSORS_CONF)
-PAGE_SIZE := $(shell getconf PAGESIZE)
+# TOTAL_CORES := $(shell getconf _NPROCESSORS_CONF)
+# PAGE_SIZE := $(shell getconf PAGESIZE)
```
bushidocodes, line 9: Can you please uncomment this? I don't see this defined anywhere else, and we do use this for guard pages and such.

For example:
```c
void *addr = mmap(NULL, sandbox_size + linear_memory_max_size + /* guard page */ PAGE_SIZE, PROT_NONE,
	                  MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
```

runtime/Makefile
```diff
@@ -5,8 +5,8 @@ $(error Unsupported Architecture. Supports x86_64 and aarch64, found $(ARCH))
 endif
 endif
 
-TOTAL_CORES := $(shell getconf _NPROCESSORS_CONF)
-PAGE_SIZE := $(shell getconf PAGESIZE)
+# TOTAL_CORES := $(shell getconf _NPROCESSORS_CONF)
```
bushidocodes, line 8: If this is going to be dynamic, can you please just delete this line?

runtime/Makefile
```diff
@@ -81,8 +81,8 @@ BINARY_NAME=sledgert
 
 # Sets a flag equal to the processor architecture
 CFLAGS += -D${ARCH}
-CFLAGS += -DNCORES=${TOTAL_CORES}
-CFLAGS += -DPAGE_SIZE=$(PAGE_SIZE)
+# CFLAGS += -DNCORES=${TOTAL_CORES}
+# CFLAGS += -DPAGE_SIZE=$(PAGE_SIZE)
```
bushidocodes, line 85: Also uncomment this to declare the macro

runtime/Makefile
```diff
@@ -81,8 +81,8 @@ BINARY_NAME=sledgert
 
 # Sets a flag equal to the processor architecture
 CFLAGS += -D${ARCH}
-CFLAGS += -DNCORES=${TOTAL_CORES}
-CFLAGS += -DPAGE_SIZE=$(PAGE_SIZE)
+# CFLAGS += -DNCORES=${TOTAL_CORES}
```
bushidocodes, line 84: and delete this

runtime/experiments/applications/ekf/one_iteration/spec.json
```diff
@@ -1,6 +1,6 @@
 {
   "active": true,
-  "name": "resize",
+  "name": "ekf",
```
bushidocodes, line 3: Good catch!

runtime/src/main.c
```diff
@@ -259,10 +259,9 @@ log_compiletime_config()
 	printf("\tArchitecture: %s\n", "x86_64");
 #endif
 
-	int ncores = NCORES;
-	printf("\tTotal Cores: %d\n", ncores);
-	int page_size = PAGE_SIZE;
-	printf("\tPage Size: %d\n", page_size);
+#ifdef PAGE_SIZE
```
bushidocodes, line 262: I think this might be required. If so, this probably shouldn't be in a conditional. You want also want an `#ifndef` that triggers a `#warn` if not set as well.

runtime/src/runtime.c
```diff
@@ -91,6 +95,11 @@ runtime_set_resource_limits_to_max()
 void
 runtime_initialize(void)
 {
+	runtime_worker_threads          = malloc(runtime_worker_threads_count * sizeof(pthread_t));
```
bushidocodes, line 98: Any reason not to use `calloc` here as well? I believe there is an assert like the following elsewhere:

```c
assert(runtime_worker_threads[i] != 0);
```

runtime/src/runtime.c
```diff
@@ -42,6 +42,10 @@ runtime_cleanup()
 {
 	if (runtime_sandbox_perf_log != NULL) fflush(runtime_sandbox_perf_log);
 
+	if (runtime_worker_threads_deadline) free(runtime_worker_threads_deadline);
```
bushidocodes, line 45: Nice work finding the right place to hook into for this!

runtime/src/software_interrupt.c
```diff
@@ -47,13 +47,14 @@ software_interrupt_deferred_sigalrm_max_print()
 		printf("Worker %d: %d\n", i, software_interrupt_deferred_sigalrm_max[i]);
 	}
 	fflush(stdout);
+	if (software_interrupt_deferred_sigalrm_max) free((void *)software_interrupt_deferred_sigalrm_max);
```
bushidocodes, line 50: Kinda weird to use print to free. Maybe create `software_interrupt_cleanup` and call that after `software_interrupt_deferred_sigalrm_max_print` in `runtime_cleanup`?

runtime/src/runtime.c
bushidocodes: Ahh, I see you removed that below. Was there a specific reason?

bushidocodes (changes requested):
Generally looks good. I've left some general cleanup suggestions and would like some feedback on some parts. Spinning up Docker now to run and will leave comments if I hit any additional surprises.

[Edit after running]: No additional comments. Everything compiled and the handful of tests I ran seemed to execute as expected.

runtime/Makefile
emil916: It is being defined in the `types.h` file. Since @gparmer  said it should always be **4096**, I figured reading it from the machine is unnecessary.

bushidocodes: Cool. Just go ahead and delete these zombie lines from the makefile then! 🙌

runtime/src/runtime.c
emil916: I removed it because at this point the thread ids cannot be uninitialized, since we check successful `pthread_create` right after each create.

runtime/src/software_interrupt.c
emil916: Yeah, I hesitated whether I should create a separate function for just one call. But agreed, this is not good SE. So, I'll do what you said.

runtime/src/runtime.c
gparmer: Assertions are there to validate and document assumptions. It is rare to make an assumption and use that as a justification to remove an assumption.

runtime/src/software_interrupt.c
gparmer: The cast to `(void *)` is quite strange to me. Is it required to remove a warning? If it is superfluous, then I'd remove it. I don't think that it helps document the use, and it is far enough away from my sense of typical C that I've seen that it feels like an outlier.

runtime/src/software_interrupt.c
```diff
@@ -72,9 +79,6 @@ sigalrm_propagate_workers(siginfo_t *signal_info)
 		for (int i = 0; i < runtime_worker_threads_count; i++) {
 			if (pthread_self() == runtime_worker_threads[i]) continue;
 
-			/* All threads should have been initialized */
-			assert(runtime_worker_threads[i] != 0);
```
gparmer, line 76: This assertion makes sense to me, but I would have it before line 73. If the assertion is on a value, we should trigger the assertion *before* we use the value.

gparmer (approved):
Looks fantastic! I don't have many serious comments. There's a gap in my knowledge as to where `runtime_worker_threads_count` is set, so I cannot comment on correctness WRT that.

runtime/src/software_interrupt.c
emil916: Yes, I did it to remove the warning. I think it is because of its type that is unusual (volatile _Atomic(sig_atomic_t) ). Any other better way to do it without casting?

runtime/Makefile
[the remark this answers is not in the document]
emil916: I removed it below in the `CFLAGS`, I forgot to remove them on top. I'll do now.
So, I believe ((@bushidocodes please correct me if I'm wrong) keeping it in the `CFLAGS` (with `-D` ) eventually is the same thing as `#define` in the code. So,, right now it is defined in the `types.h` file as constant 4096.

runtime/Makefile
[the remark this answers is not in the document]
emil916: This is not supposed to be there, I removed it already yesterday after Sean's review. The latest commit does not have this.

runtime/src/software_interrupt.c
emil916: Okay, that makes sense. Putting it before that line.

runtime/src/runtime.c
emil916: Agreed. Changing `malloc` to `calloc`.

PR #262, opened 2021-07-23
linear chain function support

lyuxiaosu:
Please ignore the newly added scripts or function code, or Makefile and other scripts changed, these won't be merged to the master branch.

Please only focus on the sledge code. This version includes the following implementation:
1. Support linear chain function calling
2. Correct the queuing time and total time for each sandbox in the chain.


runtime/src/current_sandbox.c
```diff
@@ -88,20 +103,43 @@ current_sandbox_start(void)
...
+                                  sandbox_request_allocate(next_module, false, sandbox->request_length, 
+							   next_module->name, sandbox->client_socket_descriptor,
+                                                           (const struct sockaddr *)&sandbox->client_address,
+                                                           sandbox->request_arrival_timestamp, enqueue_timestamp, 
+							   true, pre_func_output, output_length);
+		/* TODO: all sandboxs in the chain share the same request id, but sandbox_request_allocate() will busy-wait to generate an unique
+		   id, should we optimize it here?*/
+		sandbox_request->id = sandbox->id;  
+		/* Add to the Global Sandbox Request Scheduler */
+		global_request_scheduler_add(sandbox_request);
```
gparmer, line 123: I'm guessing that this code does *not* include your optimization for avoiding the global queue in the case that we can run the next sandbox on our local core.

@bushidocodes @emil916 Should we expect any problems with workers adding onto the global queue in *addition* to the listener core?

runtime/src/current_sandbox.c
```diff
@@ -88,20 +103,43 @@ current_sandbox_start(void)
...
+		if (sandbox->request_from_outside) {
+			sandbox_remove_from_epoll(sandbox);
```
gparmer, line 125: I think that I'm missing something there.

As we never returned from `accept` to trigger this sandbox, we didn't create a new `fd` for it. So that was never added to the `epoll` set. Is this just removing the initial `fd` used to access the sandbox from the network from the epoll set? If so, then we should *not* be doing it for each activation of the sandbox (this is not just a performance problem, you could be removing `fd`s that belong to other sandboxes). So if this is removing the `fd` for the overall module, then I'd expect it would be done 1. at bootup time when we're parsing the modules, and 2. should also include closing the fd. Thus, it would be done only once.

I get the feeling here that I'm missing something. I think this is an important discussion. @bushidocodes @emil916

runtime/src/global_request_scheduler_minheap.c
```diff
@@ -19,7 +19,8 @@ global_request_scheduler_minheap_add(void *sandbox_request)
 {
 	assert(sandbox_request);
 	assert(global_request_scheduler_minheap);
-	if (unlikely(!listener_thread_is_running())) panic("%s is only callable by the listener thread\n", __func__);
+	/* panic is not applicable here because all worker_threads can add a request to the global queue if it's a chain function */
+	//if (unlikely(!listener_thread_is_running())) panic("%s is only callable by the listener thread\n", __func__);
```
gparmer, line 23: I think that the check is still correct (how would a worker be running without the listener?), but the printout is very wrong. Something more appropriate would be `"%s is only callable after the listener thread has begun execution"`. I don't know if it adds much value as a check, though.

runtime/src/global_request_scheduler_minheap.c
```diff
@@ -74,8 +75,7 @@ sandbox_request_get_priority_fn(void *element)
  * Initializes the variant and registers against the polymorphic interface
  */
 void
-global_request_scheduler_minheap_initialize()
-{
+global_request_scheduler_minheap_initialize() {
```
gparmer, line 78: Breaks style.

runtime/src/current_sandbox.c
gparmer: Check return value? I'm guessing this is trying to remove `0` from the epoll set, which should error out?

runtime/src/module.c
```diff
@@ -374,6 +374,7 @@ module_new_from_json(char *file_name)
 	int   module_count    = 0;
 	char *request_headers = NULL;
 	char *reponse_headers = NULL;
+	struct module *tail_module = NULL;
```
gparmer, line 377: vertically line up initializations.

runtime/src/module.c
```diff
@@ -555,8 +556,18 @@ module_new_from_json(char *file_name)
 			                                   relative_deadline_us, port, request_size, response_size,
 			                                   admissions_percentile, expected_execution_us);
 			if (module == NULL) goto module_new_err;
-
+			
 			assert(module);
+
+			if (tail_module == NULL) {
```
gparmer, line 562: We'll need to come up with a json representation for this, so I know this isn't the "final" form of this code. However, I wanted to point out that these conditionals are perhaps a little more complex than they have to be. I believe you can replace the conditionals with:

```c
module->next_module      = NULL;
tail_module->next_module = module;
tail_module              = module;
```

gparmer (commented):
I'm very impressed that you were able to make these changes relatively minimal. It looks really good. My comments are mostly relatively minimal. My only real concern is in the function to remove from the `epoll` set. However, I think that you really want @bushidocodes @emil916 to have a look to make sure I'm not missing some assumptions might be broken here.

I think that this code does *not* include your optimization to avoid using the global queue in some cases.

runtime/include/module.h
```diff
@@ -78,6 +78,7 @@ struct module {
 
 	/* Entry Function to invoke serverless function */
 	mod_main_fn_t main;
+	struct module *next_module; /* the next module in the chain */
```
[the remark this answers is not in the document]
lyuxiaosu: Yes, I think so.

runtime/include/sandbox_functions.h
```diff
@@ -28,6 +28,19 @@ sandbox_close_http(struct sandbox *sandbox)
...
 
+/**
+ * Remove the client fd from epoll
+ */
+static inline void
+sandbox_remove_from_epoll(struct sandbox *sandbox)
+{
+        assert(sandbox != NULL);
+
+        int rc = epoll_ctl(worker_thread_epoll_file_descriptor, EPOLL_CTL_DEL, sandbox->client_socket_descriptor, NULL);
```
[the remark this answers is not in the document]
lyuxiaosu: You are right. The code does the same as you said. If the module is in the middle of the chain, it shouldn't be monitored by the epoll, so simply just not adding it to the epoll set. If the module is the first or last module, it will be added to the epoll. For the first module, the client fd should be removed from the epoll when it is finished, but won't be closed. The original sledge will remove the fd from the epoll and close it. For the chain function, we won't close it in the first module but will do in the last module, so we need to differentiate a module is a first module, or a middle module or the last module.

runtime/include/sandbox_functions.h
```diff
@@ -28,6 +28,19 @@ sandbox_close_http(struct sandbox *sandbox)
 	client_socket_close(sandbox->client_socket_descriptor, &sandbox->client_address);
 }
 
+/**
+ * Remove the client fd from epoll
+ */
+static inline void
```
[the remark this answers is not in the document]
lyuxiaosu: I don't understand the code performance for the fastpath here. My understanding for the inline is it won't jump to the code but it has the whole block of the code in there because the code has been copied. If it can be inline, the compiler will do it; if it cannot, the compiler will ignore it. I add `inline` because I saw all functions in this file has inline, but I don't know how it is related to the fastpath, could you explain more about the fastpath?

runtime/include/sandbox_request.h
```diff
@@ -15,13 +15,17 @@
...
+	bool		request_from_outside; /* true is yes, false is no */
 	struct module * module;
 	char *          arguments;
 	int             socket_descriptor;
 	struct sockaddr socket_address;
 	uint64_t        request_arrival_timestamp; /* cycles */
+	uint64_t        enqueue_timestamp; /* cycles */
 	uint64_t        absolute_deadline;         /* cycles */
-
+	char * previous_function_output;
```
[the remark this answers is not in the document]
lyuxiaosu: ah, got it, will modify it.

runtime/include/sandbox_request.h
```diff
@@ -15,13 +15,17 @@
...
+	ssize_t output_length;
+	ssize_t previous_request_length;                    /* previous request length */
```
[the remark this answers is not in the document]
lyuxiaosu: yes, will remove it

runtime/src/current_sandbox.c
lyuxiaosu: > I think that I'm missing something there.
> 
> As we never returned from `accept` to trigger this sandbox, we didn't create a new `fd` for it. So that was never added to the `epoll` set. Is this just removing the initial `fd` used to access the sandbox from the network from the epoll set? If so, then we should _not_ be doing it for each activation of the sandbox (this is not just a performance problem, you could be removing `fd`s that belong to other sandboxes). So if this is removing the `fd` for the overall module, then I'd expect it would be done 1. at bootup time when we're parsing the modules, and 2. should also include closing the fd. Thus, it would be done only once.
> 
> I get the feeling here that I'm missing something. I think this is an important discussion. @bushidocodes @emil916

You are right, the return of accept won't trigger the sandbox. The fd here is the return value of accept(), it is the client fd. When a sandbox is running, it will call "sandbox_open_http", within this function, it will add this client fd to the epoll to monitor its IN/OUT/ERR event because sledge will do read/write from/to this fd and it can be blocked. If it is the first module (sandbox->request_from_outside is true if it is the first module), sledge should remove the client fd from epoll but not close if before the sandbox exits, otherwise, the sandbox is freed and epoll still return the client fd, but epoll_event.data.ptr is invalid because the sandbox is freed(epoll_event.data.ptr = (void *)sandbox), so we must remove it here when the sandbox finished.

runtime/src/global_request_scheduler_minheap.c
lyuxiaosu: This `global_request_scheduler_minheap_add` is adding a request to the global queue. In the original sledge, only the listener thread creates the request and adds it to the global queue, so `listener_thread_is_running()` won't panic. Right now, it's not only the listener thread can add a request to the queue, but all the worker threads can do that, so here `listener_thread_is_running()` will panic because the current thread might not be the listener thread.

`listener_thread_is_running()` checks if the current thread is the listener thread or not. Sledge will panic here if not comment on this line.

runtime/src/global_request_scheduler_minheap.c
lyuxiaosu: Sorry, I think I changed the breaks style when I traced the code, the original sledge code wasn't like this. I will recover it.

runtime/src/module.c
lyuxiaosu: got it, will modify it.

runtime/src/module.c
lyuxiaosu: but tail_module is initialized to NULL, no allocated memory, tail_module->next_module is not accessible.

runtime/src/current_sandbox.c
lyuxiaosu: > Check return value? I'm guessing this is trying to remove `0` from the epoll set, which should error out?

Yes, should check the error. It's inside of `the sandbox_remove_from_epoll `
```
static inline void
sandbox_remove_from_epoll(struct sandbox *sandbox)
{
        assert(sandbox != NULL);

        int rc = epoll_ctl(worker_thread_epoll_file_descriptor, EPOLL_CTL_DEL, sandbox->client_socket_descriptor, NULL);
        if (unlikely(rc < 0)) panic_err();

}
```

runtime/include/sandbox_request.h
```diff
@@ -65,9 +69,10 @@ sandbox_request_log_allocation(struct sandbox_request *sandbox_request)
  * @return the new sandbox request
  */
 static inline struct sandbox_request *
-sandbox_request_allocate(struct module *module, char *arguments, int socket_descriptor,
-                         const struct sockaddr *socket_address, uint64_t request_arrival_timestamp,
-                         uint64_t admissions_estimate)
+sandbox_request_allocate(struct module *module, bool request_from_outside, ssize_t request_length,
```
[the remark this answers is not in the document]
lyuxiaosu: OK, the parameters in this function are indeed too many.

runtime/include/sandbox_request.h
```diff
@@ -76,11 +81,16 @@ sandbox_request_allocate(struct module *module, char *arguments, int socket_desc
 	sandbox_request->id = sandbox_request_count_postfix_increment();
 
 	sandbox_request->module            = module;
+	sandbox_request->request_from_outside = request_from_outside;
```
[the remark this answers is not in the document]
lyuxiaosu: I forgot to use the `format.sh` to line up the code.

runtime/include/worker_thread_execute_epoll_loop.h
```diff
@@ -24,7 +24,6 @@ worker_thread_execute_epoll_loop(void)
 		struct epoll_event epoll_events[RUNTIME_MAX_EPOLL_EVENTS];
 		int                descriptor_count = epoll_wait(worker_thread_epoll_file_descriptor, epoll_events,
                                                   RUNTIME_MAX_EPOLL_EVENTS, 0);
-
 		if (descriptor_count < 0) {
```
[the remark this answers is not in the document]
lyuxiaosu: I think I deleted this line when I cleaned up some comments I added, will recover it.

runtime/include/sandbox_functions.h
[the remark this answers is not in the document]
gparmer: My concern here is that it seems that this is being done as part of the inter-sandbox communication, right? We want to do any sandbox/chain setup at the time we add the modules.

runtime/include/sandbox_functions.h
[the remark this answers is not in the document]
gparmer: The function likely shouldn't be in this file, and instead in a .c file with the rest of the functions that aren't as performance sensitive.

runtime/src/current_sandbox.c
```diff
@@ -67,13 +68,27 @@ current_sandbox_start(void)
 	char *error_message = "";
 
 	sandbox_initialize_stdio(sandbox);
+	struct module * next_module = sandbox->module->next_module;
 
-	sandbox_open_http(sandbox);
+	if (sandbox->request_from_outside || next_module == NULL) {
```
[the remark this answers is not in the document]
lyuxiaosu: Will add comment for this line. The first and last sandbox in the chain should use epoll to monitor the fd because they needs to read and write from/to this fd. `sandbox->request_from_outside` indicates the first sandbox in the chain, `next_module == NULL` indicates the last sandbox in the chain.

runtime/src/current_sandbox.c
```diff
@@ -88,20 +103,43 @@ current_sandbox_start(void)
...
 	sandbox->completion_timestamp = __getcycles();
 
-	/* Retrieve the result, construct the HTTP response, and send to client */
-	if (sandbox_send_response(sandbox) < 0) {
-		error_message = "Unable to build and send client response\n";
-		goto err;
-	};
 
-	http_total_increment_2xx();
+	if (next_module != NULL) {
```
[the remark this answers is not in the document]
lyuxiaosu: > To help readers understand what this semantically means, I wonder if providing a (`static inline`) function that associates a useful name with these checks would be useful. For example `bool sandbox_output_to_network(sandbox)`. If you think this is a good idea, then you'd also likely have a `sandbox_input_from_network` as well.

I think the better idea is to define an enum for the sandbox type. We need to mark the type of the sandbox. For the linear chain function, there are three types of the sandbox - FIRST, MIDDLE, LAST, but the name might not be the best.

runtime/src/current_sandbox.c
gparmer: Thanks for the input! I was confused, and thought the epoll delete was done during IPC, but you're right: it is done at initialization. I want to confirm that these goals are maintained:

- no sockets are created for modules that are not outward facing.
- Only outward-facing modules have sockets, and `accept` returns a *single* client's fd.
- the socket used to communicate with the client is both used to read the input, and to send the output *after* the last sandbox in the chain executes.

I'm hesitant to believe that these goals are met as it seems that the fds are made, added, then removed.

runtime/src/global_request_scheduler_minheap.c
gparmer: I understand all of this. I think that I don't understand what `listener_thread_is_running` does. Given your response, it really tells us if the listener thread is currently the thread that is executing this thread, I assume. In that case, I understand your response. I thought it simply returned if the thread had been created or not. Thanks for the clarification.

runtime/src/current_sandbox.c
```diff
@@ -88,20 +103,43 @@ current_sandbox_start(void)
...
+		/* Generate a new request, copy the current sandbox's output to the next request's buffer, and put it to the global queue */
+		ssize_t output_length = sandbox->request_response_data_length - sandbox->request_length;
+		char * pre_func_output = (char *) malloc(output_length);
+		memcpy(pre_func_output, sandbox->request_response_data + sandbox->request_length, output_length);
```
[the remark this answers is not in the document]
lyuxiaosu: > I don't follow the logic here with the `+ request_length`. Does the sledge code (before your addition) use the same buffer for holding the request _and_ the response? Sorry for forgetting here!

Yes, this is also my confusion when I saw sledge code at the first time. This is the original code design. It use the same buffer for holding the request and response. See the name `request_response_data`.  Also see the comment of the function to prepare http response.
```
static inline int
sandbox_send_response(struct sandbox *sandbox)
{
        assert(sandbox != NULL);

        /*
         * At this point the HTTP Request has filled the buffer up to request_length, after which
         * the STDOUT of the sandbox has been appended. We assume that our HTTP Response header is
         * smaller than the HTTP Request header, which allows us to use memmove once without copying
         * to an intermediate buffer.
         */
        memset(sandbox->request_response_data, 0, sandbox->request_length);
```

runtime/src/module.c
gparmer: Thanks, I figured I missed a case. Maybe focus on only that difference, keeping most code shared, then?

```c
if (tail_module != NULL) tail_module->next_module = module;
tail_module->next_module = NULL;
tail_module              = module;
```
Point being: I had to read this code about 5 times before I understood it. And I still got the reformat wrong ;-) It feels like it can be simplified.

runtime/src/current_sandbox.c
```diff
@@ -88,20 +103,43 @@ current_sandbox_start(void)
...
+	 	uint64_t enqueue_timestamp = __getcycles();	
+		struct sandbox_request *sandbox_request =
+                                  sandbox_request_allocate(next_module, false, sandbox->request_length, 
+							   next_module->name, sandbox->client_socket_descriptor,
+                                                           (const struct sockaddr *)&sandbox->client_address,
+                                                           sandbox->request_arrival_timestamp, enqueue_timestamp, 
+							   true, pre_func_output, output_length);
+		/* TODO: all sandboxs in the chain share the same request id, but sandbox_request_allocate() will busy-wait to generate an unique
+		   id, should we optimize it here?*/
```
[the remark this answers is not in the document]
lyuxiaosu: > It uses `faa` to retrieve the new id, right? In that case, there should be no busy waiting.
> 
> If it doesn't use `faa`, we should certainly discuss optimizations.
> 
> Separately, it would be good to get rid of the `faa` on global memory. Easy enough to have the sandbox ids be retrieved with a `faa` on local (`__thread`) memory bitwise appended onto the worker id. @bushidocodes Is this insane? Should we create a card? Certainly, this is very low priority.

what is `faa` ? sledge get this id by call C function `atomic_fetch_add`, it applies for the variable defined by `_Atomic` keyword. I am not sure the compiler will use the software or hardware to guarantee the atomic operation.

runtime/src/current_sandbox.c
lyuxiaosu: > I'm guessing that this code does _not_ include your optimization for avoiding the global queue in the case that we can run the next sandbox on our local core.
> 
> @bushidocodes @emil916 Should we expect any problems with workers adding onto the global queue in _addition_ to the listener core?

Yes, this code doesn't include the optimization part for avoiding the global queue, I still have some concerns for this but I will define a macro to control this optimization.

runtime/src/current_sandbox.c
gparmer: I should also note: if the fd is created for modules in a chain, currently, that's likely OK at the current point in time. Once you add the json support for chains to form them along-side other modules, then it will be a good time to fix which modules have external sockets and which dont.

runtime/include/sandbox_functions.h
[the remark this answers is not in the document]
lyuxiaosu: > My concern here is that it seems that this is being done as part of the inter-sandbox communication, right? We want to do any sandbox/chain setup at the time we add the modules.

Yes, you are right. When a module was added, it will create the server side socket fd and add it to epoll. The fd here is the client fd - the return value of the `accept()`. Sledge also add this fd to epoll to monitor its IN/OUT/ERR event because sledge will read/write data from/to this fd, which can be blocked, but this read/write operation is done within a sandbox. The client fd cannot be created when a module was added.

runtime/include/sandbox_functions.h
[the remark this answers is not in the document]
lyuxiaosu: I agree. I don't know why these epoll operation was set an `inline` function. There is no corresponding .c file for this header file. If this function should be put into a .c file, I think the other functions in this file, such as `sandbox_open_http` and `sandbox_close_http` should also be moved into the same .c file.

runtime/src/current_sandbox.c
lyuxiaosu: Modules only create and epoll monitor the socket fds for the server-side(listen the socket fd to wait for the requests coming in), this is done when a module was added. The listening thread will do `accept()` and return a client-side fd, this client fd is monitored by epoll, read/write data within a sandbox.

runtime/src/module.c
lyuxiaosu: > Thanks, I figured I missed a case. Maybe focus on only that difference, keeping most code shared, then?
> 
> ```c
> if (tail_module != NULL) tail_module->next_module = module;
> tail_module              = module;
> tail_module->next_module = NULL;
> ```
> 
> Point being: I had to read this code about 5 times before I understood it. And I still got the reformat wrong ;-) It feels like it can be simplified.

I totally agree, this is more clear/readable than my version, I will change the code like this, sorry for the 5 times reading:)

runtime/src/current_sandbox.c
emil916: I believe we should not expect any problems, as long as we have LOCKs enabled to access the global queue, then workers technically can add to it with no problem.
I don't get line 136 much, is this for making sure the 2nd and the 3rd inner functions of the chain get handled by the same worker? But even if so, how will the different sandboxes have the same exact absolute deadlines? I must be missing something here.

runtime/include/sandbox_request.h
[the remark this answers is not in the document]
emil916: Yes, you can use either malloc/free memory or to avoid dynamic mem alloc, then just create a local struct instance like this:
```
struct sandbox_chained_input sci;
sci.request_length = request_length;
...
```

runtime/src/current_sandbox.c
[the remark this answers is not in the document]
emil916: @lyuxiaosu  is right, the Sledge code does use the same buffer for both the request and the response. My only comment about the above code, I don't see why creating an extra variable for the output is really needed (+malloc). The `responspe_data `with the proper output cursor would be sufficient.

runtime/src/global_request_scheduler_minheap.c
emil916: The name of the this function is indeed confusing. Should be something like: `is_this_listener_thread()`

runtime/src/current_sandbox.c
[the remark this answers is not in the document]
bushidocodes: Going through all this for the first time now, but I suspect that generalizing the sandbox / sandbox request structures as a generic type (first member enum followed by a union) is a good approach, especially if you want to compare different implementations.

.gitmodules
```diff
@@ -1,6 +1,6 @@
 [submodule "awsm"]
 path = awsm
-url = https://github.com/gwsystems/aWsm
+url = https://github.com/lyuxiaosu/aWsm.git
```
bushidocodes, line 3: I don't see any changes in your fork, so please be sure to revert this to gwsystems.

add_partition.sh
```diff
@@ -0,0 +1,55 @@
+#/bin/bash
```
bushidocodes, line 1: Nice! A bash script! 🙌 I suggest using shellcheck and sh on this.
https://github.com/mvdan/sh
https://www.shellcheck.net/#

If possible, please also add a comment block at the top explaining what this is for. I assume this is for CloudLab setup stuff?

install_docker.sh
```diff
@@ -0,0 +1,8 @@
+sudo apt-get update
```
bushidocodes, line 1: Can you please add a comment here about how this is used? I'm guessing this is run on the baremetal host and `add_partition.sh` runs inside the docker container?

runtime/include/sandbox_request.h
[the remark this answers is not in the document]
bushidocodes: To be clear, C "complex" types like structs are just pass by value like other types. It's not like a class in a high level language where it has to be on the heap. Here's a small REPL to show some of the implications of that! https://replit.com/@spmcbride1201/pass-options

runtime/src/current_sandbox.c
bushidocodes: I wrote the global request queue minheap variant specifically assuming that it's only called by the listener and panics out the program if it's called by a worker. I assume that was changed in this PR for this to function.

runtime/src/global_request_scheduler_minheap.c
bushidocodes: The original intent here was to panic if a worker called this function because I had perf concerns about multiple threads adding to the global structure.

runtime/include/sandbox_request.h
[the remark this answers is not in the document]
lyuxiaosu: My concern for using malloc/free is the performance issue, I tend to use static variables that stay in the stack, that's @emil916 's example code. If this struct has some pointer (pointing to the dynamic memory) and is passed by value, I think the pointer is also copied and points to the same dynamic memory, that is fine, I think.

runtime/src/current_sandbox.c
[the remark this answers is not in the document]
lyuxiaosu: Are you talking about the type of the sandbox? Yes, for the linear chain function, it's better to define an enum to describe the type of a sandbox, FIRST, MIDDLE, and LAST, then based on the type to decide which block of code to execute.

runtime/src/current_sandbox.c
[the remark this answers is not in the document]
lyuxiaosu: I cannot find `responspe_data`, are you mentioning `sandbox->request_response_data`? The malloc here is for copying the currency sandbox output to the next sandbox's request buffer, when the next sandbox running, it won't read the input from the networking if it is the middle sandbox in the chain, it will read the input from this malloced buffer, otherwise, when the current sandbox is freed, its `request_response_data` is freed.

runtime/src/current_sandbox.c
lyuxiaosu: > I believe we should not expect any problems, as long as we have LOCKs enabled to access the global queue, then workers technically can add to it with no problem.
> I don't get line 136 much, is this for making sure the 2nd and the 3rd inner functions of the chain get handled by the same worker? But even if so, how will the different sandboxes have the same exact absolute deadlines? I must be missing something here.

You are right, line 136 tries to check if the current sandbox has the same deadline with the next sandbox request or not, if they are the same, the worker thread will put the next sandbox(after initializing it) into its local runqueue not the global queue, this is to avoid the global queue overhead. But based on my performance tests, this optimization seems not to improve the performance, so I add a macro to control this feature, the default setting for this feature is off.

.gitmodules
lyuxiaosu: > I don't see any changes in your fork, so please be sure to revert this to gwsystems.

Good catch, I didn't change anything for aWsm. The original thought was not want the update of the compiler to affect my branch:), finally, this won't be submitted.

add_partition.sh
lyuxiaosu: right, this is for extending the partition for cloudlab machine because there is only 16G for / path in cloudlab machine, very easy to exhaust the disk space if you create many docker containers, we usually extend the / space when we use cloudlab machine.

install_docker.sh
lyuxiaosu: Yes, `add_partition.sh` and `install_docker.sh` should be used together. `add_partition.sh` is for extending the space of / path in cloudlab machines. Then run `install_docker.sh` to install docker and let the docker to write all its images and other stuff to this new parttion.

runtime/include/perf_window.h
```diff
@@ -158,3 +162,31 @@ perf_window_get_count(struct perf_window *self)
...
+	if (self->count % PERF_WINDOW_BUFFER_SIZE != 0) { return; }
+	/* Not need to hold lock because this operation won't add/delete the array */
+	float min = self->by_duration[0].execution_time/1000.0;
+	float max = self->by_duration[PERF_WINDOW_BUFFER_SIZE-1].execution_time/1000.0;
+	uint64_t sum = 0;
+        float fifty_p = self->by_duration[PERF_WINDOW_BUFFER_SIZE * 50 / 100].execution_time/1000.0;	
+        float seventy_p = self->by_duration[PERF_WINDOW_BUFFER_SIZE * 70 / 100].execution_time/1000.0;	
+        float eighty_p = self->by_duration[PERF_WINDOW_BUFFER_SIZE * 80 / 100].execution_time/1000.0;	
+        float nighty_p = self->by_duration[PERF_WINDOW_BUFFER_SIZE * 90 / 100].execution_time/1000.0;	
+        float nighty_night_p = self->by_duration[PERF_WINDOW_BUFFER_SIZE * 99 / 100].execution_time/1000.0;
```
emil916, line 182: *ninety and *ninety_nine ?

runtime/include/sandbox_functions.h
```diff
@@ -215,6 +215,16 @@ sandbox_mem_print_perf(struct sandbox *sandbox)
         uint32_t blocked_us      = sandbox->blocked_duration / runtime_processor_speed_MHz;
         uint32_t returned_us     = sandbox->returned_duration / runtime_processor_speed_MHz;
 
+	if (sandbox->module->next_module == NULL) {
+                uint32_t now = __getcycles();
+                bool miss_deadline = now > sandbox->absolute_deadline ? true : false;
```
emil916, line 220: I may be missing something here (like when this will be called), but in any case I think instead of "now" using sandbox's `completion_timestamp` would be more accurate.

runtime/include/perf_window.h
lyuxiaosu: ah, I guess you suggest to rename the variables here, right? I am sorry the name is very confused, "p" does not mean pointer, it means percentile.

runtime/include/sandbox_functions.h
lyuxiaosu: You are right, thanks for catching this potential bug. This is for evaluating the performance of the scheduling algrithms, not related to any functionalities of sledge.

PR #264, opened 2021-08-09
WIP: Multi-Tenancy first pass

emil916:


runtime/src/global_request_scheduler_mts.c
```diff
@@ -0,0 +1,333 @@
...
+		panic("Tried to delete a non-present MGRQ from the Global Default queue. Already deleted?, ITS SIZE: "
+		      "%d",
+		      priority_queue_length_nolock(mgrq->sandbox_requests));
+	}
+
+	/* Add the corresponding MGRQ to the Default queue */
+	rc = priority_queue_enqueue_nolock(global_request_scheduler_mts_guaranteed, mgrq);
+	if (rc == -ENOSPC) panic("Global Guaranteed queue is full!\n");
+
+done:;
```
gparmer, line 55: stray `;`?

runtime/src/global_request_scheduler_mts.c
```diff
@@ -0,0 +1,333 @@
...
+ * @param mt_class the multi-tenancy class of the global request to compare the target deadline against
+ * @returns 0 if successful, -ENOENT if empty or if request isn't earlier than target_deadline
+ */
+int
+global_request_scheduler_mts_remove_with_mt_class(struct sandbox_request **removed_sandbox_request,
+                                                          uint64_t                 target_deadline,
+                                                          enum MULTI_TENANCY_CLASS target_mt_class)
+{
+	/* Avoid unnessary locks when the target_deadline is tighter than the head of the Global runqueue */
+	uint64_t global_guaranteed_deadline = priority_queue_peek(global_request_scheduler_mts_guaranteed);
```
gparmer, line 166: accessing global structures without lock? likely to result in races. Want to store the deadline in a global and access that instead.

runtime/src/global_request_scheduler_mts.c
```diff
@@ -0,0 +1,333 @@
...
+	uint64_t global_default_deadline    = priority_queue_peek(global_request_scheduler_mts_default);
+
+	switch (target_mt_class) {
+	case MT_GUARANTEED:
+		if (global_guaranteed_deadline >= target_deadline) return -ENOENT;
+		break;
+	case MT_DEFAULT:
+		if (global_guaranteed_deadline == UINT64_MAX && global_default_deadline >= target_deadline)
```
gparmer, line 174: Yeah, this logic seems like it might benefit from having a function (i..e a name) associated with it as I'm seeing it again, and have to reverse engineer it each time.

runtime/src/global_request_scheduler_mts.c
```diff
@@ -0,0 +1,333 @@
...
+			return -ENOENT;
+		break;
+	}
+
+	struct module_global_request_queue *top_mgrq          = NULL;
+	struct priority_queue *             destination_queue = global_request_scheduler_mts_guaranteed;
+
+	LOCK_LOCK(&global_lock);
```
gparmer, line 182: now I understand the `_nolock` functions that were initially frightening.

runtime/src/global_request_scheduler_mts.c
```diff
@@ -0,0 +1,333 @@
...
+	}
+
+	assert(top_mgrq);
+
+	/* Remove the sandbox from the corresponding MGRQ */
+	rc = priority_queue_dequeue_nolock(top_mgrq->sandbox_requests, (void **)removed_sandbox_request);
+	if (rc == -ENOENT) goto done;
+
+	/* Delete the MGRQ from the global runqueue completely if MGRQ is empty, re-add otherwise to heapify */
+	if (priority_queue_delete_nolock(destination_queue, top_mgrq) == -1)
```
gparmer, line 210: `{...}`

runtime/src/global_request_scheduler_mts.c
```diff
@@ -0,0 +1,333 @@
...
+			rc                 = -ENOENT;
+			goto done;
+		}
+	}
+
+	assert(top_mgrq);
+
+	/* Remove the sandbox from the corresponding MGRQ */
+	rc = priority_queue_dequeue_nolock(top_mgrq->sandbox_requests, (void **)removed_sandbox_request);
+	if (rc == -ENOENT) goto done;
```
gparmer, line 207: not sure how this could happen. We hold the lock and found it be top before. assertion here?

runtime/src/global_request_scheduler_mts.c
```diff
@@ -0,0 +1,333 @@
...
+}
+
+/**
+ * @param removed_sandbox_request pointer to set to removed sandbox request
+ * @param target_deadline the deadline that the request must be earlier than to dequeue
+ * @param mt_class the multi-tenancy class of the global request to compare the target deadline against
+ * @returns 0 if successful, -ENOENT if empty or if request isn't earlier than target_deadline
+ */
+int
+global_request_scheduler_mts_remove_with_mt_class(struct sandbox_request **removed_sandbox_request,
```
gparmer, line 161: I can't really grok the logic in this function. It is very complex, and it is a testament to your abilities that you were able to get it to this state! If I were to proceed, I'd have to really reason through this, and potentially rewrite if there is any way to make it simpler. There might not be!!! Long story short, I think your ability to understand complexity is likely better than mine; sorry I'm not that useful here.

runtime/src/global_request_scheduler_mts.c
```diff
@@ -0,0 +1,333 @@
...
+
+done:
+	LOCK_UNLOCK(&global_lock);
+	return rc;
+}
+
+/**
+ * Peek at the priority of the highest priority task without having to take the lock
+ * Because this is a min-heap PQ, the highest priority is the lowest 64-bit integer
+ * This is used to store an absolute deadline
```
gparmer, line 227: This is a really risky invariant and I'm not sure I buy it completely. A simple global variable that stores the deadline seems easier and like something that won't have liveness issues.

runtime/src/global_request_scheduler_mts.c
```diff
@@ -0,0 +1,333 @@
...
+	priority_queue_free(global_request_scheduler_mts_guaranteed);
+	priority_queue_free(global_request_scheduler_mts_default);
+	priority_queue_free(global_module_timeout_queue);
+}
+
+
+void
+global_timeout_queue_add(struct module *module)
+{
+	struct module_timeout *mt = malloc(sizeof(struct module_timeout));
```
gparmer, line 292: You an avoid this allocation eventually by simply having the timer queue internal node be a part of the module. For now, the malloc is fine.

runtime/src/global_request_scheduler_mts.c
```diff
@@ -0,0 +1,333 @@
...
+	priority_queue_top_nolock(global_module_timeout_queue, (void **)&top_module_timeout);
+
+	if (top_module_timeout == NULL || now < top_module_timeout->timeout) return;
+
+	struct module *                     module          = top_module_timeout->module;
+	struct module_global_request_queue *mgrq_to_promote = module->mgrq_requests;
+
+	if (/*mgrq_to_promote->mt_class == MT_DEFAULT
+	    && priority_queue_length_nolock(mgrq_to_promote->sandbox_requests) > 0 &&*/
+	    module->remaining_budget <= 0) {
```
gparmer, line 319: YOu want to promote periodically. Not just periodically when the budget is expended. Might be missing somethign here.

runtime/src/global_request_scheduler_mts.c
```diff
@@ -0,0 +1,333 @@
...
+ * Checks if there are any tenant timers that run out in the GLOBAL queue,
+ *  if so promote that tenant.
+ */
+void
+global_timeout_queue_check_for_promotions()
+{
+	uint64_t now = __getcycles();
+
+	struct module_timeout *top_module_timeout = NULL;
+
```
gparmer, line 308: if this function is called while a lock is held, fine. Would be best to assert that fact. If not, this seems at risk of races?

runtime/src/local_runqueue_mts.c
```diff
@@ -0,0 +1,281 @@
...
+ * Adds a sandbox to the run queue
+ * @param sandbox
+ * @returns pointer to sandbox added
+ */
+void
+local_runqueue_mts_add(struct sandbox *sandbox)
+{
+	assert(sandbox != NULL);
+
+	struct perworker_module_sandbox_queue *pwm               = &sandbox->module->pwm_sandboxes[worker_thread_idx];
```
gparmer, line 55: Want to validate that the `pwm_sandboxes` elements are cache-line aligned.

runtime/src/local_runqueue_mts.c
```diff
@@ -0,0 +1,281 @@
...
+		if (sandbox->module->remaining_budget > 0) {
+			pwm->mt_class     = MT_GUARANTEED;
+			destination_queue = local_runqueue_mts_guaranteed;
+		}
+
+		/* Add sandbox module to the worker's timeout queue if guaranteed tenant and only if first sandbox*/
+		if (sandbox->module->replenishment_period > 0) {
+			struct module_timeout *mt = malloc(sizeof(struct module_timeout));
+
+			mt->timeout = get_next_timeout_of_module(sandbox->module->replenishment_period, __getcycles());
```
gparmer, line 76: validate this is safe to do without locks

runtime/src/local_runqueue_mts.c
```diff
@@ -0,0 +1,281 @@
...
+	struct priority_queue *                destination_queue = local_runqueue_mts_default;
+
+	uint64_t prev_pwm_deadline = priority_queue_peek(pwm->sandboxes);
+
+	/* Add the sandbox to the per-worker-module (pwm) queue */
+	int rc = priority_queue_enqueue_nolock(pwm->sandboxes, sandbox);
+	if (rc == -ENOSPC) panic("Per Worker Module queue is full!\n");
+	// debuglog("Added a sandbox to the PWM of module '%s'", sandbox->module->name);
+
+	if (priority_queue_length_nolock(pwm->sandboxes) == 1) {
```
gparmer, line 65: comment above this, something like "Previously the module was not in the local runqueue, and we are the first sandbox for it in this worker. Thus we need to add the module"...something better than that ;-)

runtime/src/local_runqueue_mts.c
```diff
@@ -0,0 +1,281 @@
...
+			mt->module  = sandbox->module;
+			priority_queue_enqueue_nolock(worker_thread_timeout_queue, mt);
+		}
+
+		rc = priority_queue_enqueue_nolock(destination_queue, pwm);
+
+		return;
+	}
+
+	if (sandbox->absolute_deadline < prev_pwm_deadline) {
```
gparmer, line 86: This never runs if the modules sandbox queue is length 1. I can't tell if that is OK or not. I *might* expect a check on the absolute deadline in the previous block of code. Not sure.

runtime/src/local_runqueue_mts.c
```diff
@@ -0,0 +1,281 @@
...
+ * @param sandbox to delete
+ */
+static void
+local_runqueue_mts_delete(struct sandbox *sandbox)
+{
+	assert(sandbox != NULL);
+	struct perworker_module_sandbox_queue *pwm = &sandbox->module->pwm_sandboxes[worker_thread_idx];
+
+	/* Delete the sandbox from the corresponding Per-Worker-Module queue */
+	if (priority_queue_delete_nolock(pwm->sandboxes, sandbox) == -1)
```
gparmer, line 113: {}

runtime/src/local_runqueue_mts.c
```diff
@@ -0,0 +1,281 @@
...
+	        destination_queue = local_runqueue_mts_guaranteed;
+	}
+	*/
+
+	/* Delete the PWM from the local runqueue completely if pwm is empty, re-add otherwise to heapify */
+	if (priority_queue_delete_nolock(destination_queue, pwm) == -1) {
+		// TODO: Apply the composite way of PQ deletion O(logn)
+		destination_queue = local_runqueue_mts_guaranteed;
+
+		if (priority_queue_delete_nolock(destination_queue, pwm) == -1)
```
gparmer, line 129: I might be missing that the standard allows multi-line conditionals without `{}`, but I really find them dangerous.

runtime/src/local_runqueue_mts.c
```diff
@@ -0,0 +1,281 @@
...
+	}
+	// debuglog("Deleted the PWM from the Worker Local runqueue - %s to Heapify", QUEUE_NAME);
+
+	/* Add the PWM back to the local runqueue if it still has other sandboxes inside */
+	if (priority_queue_length_nolock(pwm->sandboxes) > 0) {
+		priority_queue_enqueue_nolock(destination_queue, pwm);
+		// debuglog("Added the PWM back to the Worker Local runqueue - %s to Heapify", QUEUE_NAME);
+	} else if (pwm->module->replenishment_period > 0) {
+		struct module_timeout *dummy = NULL;
+		priority_queue_dequeue_nolock(worker_thread_timeout_queue, (void **)&dummy);
```
gparmer, line 141: need to free the timeout structure?

runtime/src/local_runqueue_mts.c
```diff
@@ -0,0 +1,281 @@
...
+
+	int rc = priority_queue_top_nolock(dq, (void **)&next_pwm);
+	while (rc != -ENOENT && next_pwm->module->remaining_budget <= 0) {
+		local_runqueue_mts_demote(next_pwm);
+		debuglog("Demoted '%s' locally", next_pwm->module->name);
+		next_pwm->mt_class = MT_DEFAULT;
+
+		rc = priority_queue_top_nolock(dq, (void **)&next_pwm);
+	}
+
```
gparmer, line 168: Given the rare use of double free lines, this is very pronounced. I don't think you are really trying to emphasize this transition, so I'd imagine you want a single empty line, not two.

runtime/src/local_runqueue_mts.c
```diff
@@ -0,0 +1,281 @@
...
+		dq = local_runqueue_mts_default;
+		rc = priority_queue_top_nolock(dq, (void **)&next_pwm);
+
+		if (rc == -ENOENT) return NULL;
+	}
+
+	struct sandbox *next_sandbox = NULL;
+
+	priority_queue_top_nolock(next_pwm->sandboxes, (void **)&next_sandbox);
+
```
gparmer, line 180: I'd imagine you want to remove empty line since the assertion is affine to the function. Up to you.

gparmer (commented):
This is really very impressive. I don't know enough about the code-base to know if the overall structure is sound, and only @bushidocodes can provide that context. I don't see any glaring bugs except perhaps one missing free (ignoring the current non-preemptive issue that @bushidocodes is working on).

I don't think that most of my comments are useful as I don't know the system enough. So take them as someone's stream of consciousness as they try to understand the code. As such, ignore any of my comments.

runtime/include/scheduler.h
```diff
@@ -25,11 +27,56 @@
...
+	uint64_t global_default_deadline    = global_request_scheduler_mts_default_peek();
+
+	/* Try to pull and allocate from the global queue if earlier
+	 * This will be placed at the head of the local runqueue */
+	switch (local_mt_class) {
+	case MT_GUARANTEED:
+		if (global_guaranteed_deadline >= local_deadline) goto done;
+		break;
+	case MT_DEFAULT:
+		if (global_guaranteed_deadline == UINT64_MAX && global_default_deadline >= local_deadline) goto done;
```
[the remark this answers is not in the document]
emil916: They are two different global variables: global_**_guaranteed_**_deadline and global_**_default_**_deadline

runtime/src/global_request_scheduler_mts.c
emil916: This is stored in a global. The peek function just retrieves that global.

runtime/src/global_request_scheduler_mts.c
emil916: Nice catch! Noted!

runtime/src/global_request_scheduler_mts.c
emil916: Oh wait, I see what you're saying. We gotta discuss this.

runtime/src/global_request_scheduler_mts.c
emil916: This function is NOT called while a lock is held, but it is called only by a single worker (that receives kernel sigalrm).

runtime/src/global_request_scheduler_mts.c
emil916: Nice catch! Noted and done!

runtime/src/local_runqueue_mts.c
emil916: I have this. But not sure how to validate it in this line, though. @gparmer 

```
struct perworker_module_sandbox_queue {
	struct priority_queue *  sandboxes;
	struct module *          module; 
	struct module_timeout    module_timeout;
	enum MULTI_TENANCY_CLASS mt_class; 
} __attribute__((aligned(128)));
```

runtime/src/local_runqueue_mts.c
emil916: When the queue length is one, there is no need to delete & add the module, just add (line 81).

runtime/src/global_request_scheduler_mts.c
```diff
@@ -0,0 +1,333 @@
...
+static struct priority_queue *global_module_timeout_queue;
+static lock_t                 global_lock;
+
+/* TODO: This function may not belong in this header file, but for now */
+static inline uint64_t
+module_request_queue_get_priority(void *element)
+{
+	struct module_global_request_queue *mgrq            = (struct module_global_request_queue *)element;
+	struct sandbox_request *            sandbox_request = NULL;
+	priority_queue_top_nolock(mgrq->sandbox_requests, (void **)&sandbox_request);
```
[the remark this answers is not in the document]
emil916: Lock is taken around the function before here.

PR #267, opened 2021-08-17
WIP: WASI Support

### LINKED ISSUE #1: Wasmception inflates Docker build context
The wasmception directory in the silverfish submodule seems to be ~2GB. We should investigate adding some of these things to a .dockerignore file to improve build times.

Blocked by aWsm work implementing WASI-SDK and follow up work integrating the new release of aWsm into sledge. 
No issue in aWsm for this, but PR is at https://github.com/gwsystems/aWsm/pull/19

### LINKED ISSUE #140: Integrate WASK SDK 11 enabled Awsm

bushidocodes:
Resolves #1
Resolves #140 

1. feat: Added WASI support for args (but only stubbed out with static values dummy_argc and dummy_argv) and reading/writing to STDIN/STDOUT. Tested this with fibonacci. Removed all wasmception code.
1. fix: Fix CI stale cache issue so ./runtime/compiletime/ changes properly invalidate cache
1. fix: Improve framework.sh to better detect sledgert segfaults on startup.
1. fix: Moved declaration before goto in perf_window.h. Compiler was complaining about this.
1. refactor: Cleanup of runtime/tests. See _Note on building WASI modules_ below
1. refactor:Stripped out unused k-v pairs in the JSON spec, including active, argsize, http-req-headers, http-req-content-type, and http-resp-headers. I intend to redo the argument support such that arguments can be defined in the JSON spec and passed into WASI syscalls, but this is not yet complete.
1. refactor: I cleaned up the members in the sandbox and sandbox_request structs to try to get things more organized and better show how a "sandbox" and a "sandbox_request" relates to WebAssembly concepts such as a "WebAssembly module", "WebAssembly instance", "WebAssembly store", and "WebAssembly instantiation." I anticipate further refactoring and documentation in the future to make this more clear and better align to the spec.
1. refactor: I removed untested and unused code relating to file descriptors. This will have to be handled substantially different to work with our abstract WASI interface, so there's minimal value in the existing code. I consider sandboxed fds a future enhancement since we don't have that with wasmception today anyways.
1. refactor: Related to above, created a new awsm_abi.h file that better shows the ABI of aWsm generated *.so files.
1. refactor: Removed residual "memory mode" USE_VM stuff that bled over from aWsm. I don't see value in the other memory models (MPX, segmentation, etc.) for the serverless runtime.
1. chore: Substantial rework of scripts, Dockerfiles, etc. Given the sheer amount of this, this is likely still a WIP and requires additional testing.
1. test: I deleted a bunch of zombie tests that haven't been used for 1yr+. This stuff would have required cleanup and refactoring for WASI, but didn't seem worth it.

Other notes:
1. There are various TODOs I need to revisit sprinkled in the diff
1. I anticipate that it'll be helpful to look at the wasmception syscall.c file in an old commit to get more WASI-SDK syscalls plumbed out. I nonetheless have deleted all that stuff form here to prevent confusion between wasmception and wasi. 
1. Pursuant to `Note on building WASI modules`, I have a ton of changes in the various submodules. I forked new branches called "WASI" for this purpose. In theory, these should be runtime agnostic rendering runtime specific branches like "awsm" versus "sledge" obsolete. 
1. CI and tests require more work.
1. SOD can't compile cleanly to WASI. Will require additional investigation.

Note on building WASI modules:

I ended up having to touch A LOT of code to be able to get our existing applications working with WASI. As such, I substantially restructured the way we handle tests. The idea is that we should have a runtime agnostic set of WASI sample applications containing tooling to compile to WASI-compliant *.wasm modules. These programs should have zero logic relating to aWsm or SLEdge and use standard tools like wasmtime to compare the execution of *.wasm module to native code. By doing this, we are able to validate our sample apps against external tools. I intend to break all of these sample apps into a distinct repo like `wasi-filter-programs`. We can then use `wasi-filter-programs` in aWsm, SLEdge, and any future repos (such as comparing against other runtimes). We can consume these sample apps in a few ways. First, we could use CI to build the *.wasm modules and download a tarball of those directly. Second, we could embed the repo as a submodule and create a wrapper Makefile that copies the *.wasm files and handles the compiler and runtime specific stuff. You can see the start of this pattern with my current Makefile.

Related to aWSM PR at https://github.com/gwsystems/aWsm/pull/52


runtime/include/arch/arch_context_t.h
```diff
@@ -11,4 +12,5 @@ struct arch_context {
 	reg_t                  regs[UREG_COUNT];
 	mcontext_t             mctx;
 	bool                   preemptable;
+	jmp_buf                start_buf;
```
gparmer, line 15: Nice! adding support for the exception handling out, I assume.

runtime/src/libc/wasi_impl_serverless.c
```diff
@@ -0,0 +1,1231 @@
...
+ * @param nwritten_retptr The number of bytes read.
+ * @return __WASI_ERRNO_SUCCESS, WASI_EAGAIN, WASI_EWOULDBLOCK, WASI_EBADF, WASI_EFAULT, WASI_EINTR, WASI_EIO,
+ * WASI_EISDIR, or others
+ */
+__wasi_errno_t
+__wasi_fd_read(void *wasi_context, __wasi_fd_t fd, const __wasi_iovec_t *iovs, size_t iovs_len,
+               __wasi_size_t *nwritten_retptr)
+{
+	/* Non-blocking copy on stdin */
+	if (fd == 0) {
```
gparmer, line 729: Usually you want to have your error condition in teh conditional first, then the mainline. I can understand why you didn't do it here, and it is fine, but it does make this harder to read.

gparmer (approved):
This is really awesome! @emil916 and  @lyuxiaosu will certainly be impacted by it (it will have merge conflicts). However, it does seem like a strong refactoring, and a good basis for adding more functionality as it adds a minimal wasi layer. I hope that @emil916 and @lyuxiaosu can review as well, but don't wait for them, I think unless they specify they can do it soon.

.gitmodules
```diff
@@ -2,6 +2,7 @@
 path = awsm
 url = https://github.com/gwsystems/aWsm
 ignore = dirty
+branch = sledge-compat
```
emil916, line 5: Oh another branch on awsm. Feels like we're going to need a doc for keeping track of all the branches haha

applications/Makefile
```diff
@@ -8,117 +8,76 @@ AWSMFLAGS= --inline-constant-globals --runtime-globals
 # --whole-archive causes the symbols in the listed static archive to be exported from the resulting *.so
 # https://stackoverflow.com/questions/805555/ld-linker-question-the-whole-archive-option
 CFLAGS=-O3 -flto
-LDFLAGS=-shared -fPIC -Wl,--export-dynamic,--whole-archive -L../libsledge/dist/ -lsledge -Wl,--no-whole-archive -lm
```
emil916, line 11: Why remove the shared option?

applications/Makefile
```diff
@@ -8,117 +8,76 @@ AWSMFLAGS= --inline-constant-globals --runtime-globals
...
+LDFLAGS=-fPIC -Wl,--export-dynamic,--whole-archive -L../libsledge/dist/ -lsledge -Wl,--no-whole-archive
 # LDFLAGS=-flto -fvisibility=hidden
 
+# Compiletime Components to link into *.so
+SLEDGE_BASE_DIR=../
+SLEDGE_RT_DIR=${SLEDGE_BASE_DIR}/runtime/
+SLEDGE_INSTALL_TARGET=${SLEDGE_RT_DIR}/bin/
```
emil916, line 17: INSTALL_TARGET__DIR_ maybe?

emil916: Or just SLEDGE_INSTALL_DIR? 
Actually maybe even SLEDGE_EXEC_DIR to be more direct?

install_deb.sh
```diff
@@ -0,0 +1,33 @@
...
+	pango1.0-tools \
+	wamerican
+
+wget $HEY_URL -O hey && chmod +x hey && sudo mv hey /usr/bin/hey
+
+wget $SHFMT_URL -O shfmt && chmod +x shfmt && sudo mv shfmt /usr/local/bin/shfmt
+
+wget $SHELLCHECK_URL -O shellcheck && chmod +x shellcheck && sudo mv shellcheck /usr/local/bin/shellcheck
+
+LLVM_VERSION=12
```
emil916, line 27: Maybe put this on top of the file?

libsledge/src/memory_instructions.c
```diff
@@ -12,55 +12,55 @@ extern thread_local struct sledge_abi__wasm_module_instance sledge_abi__current_
 INLINE uint32_t
 instruction_memory_size()
 {
-	return (uint32_t)(sledge_abi__current_wasm_module_instance.memory.size / WASM_PAGE_SIZE);
+	return (uint32_t)(sledge_abi__current_wasm_module_instance.memory.size / (uint64_t)WASM_PAGE_SIZE);
```
emil916, line 15: why the cast (uint64_t) ?  If the final return value is (uint32_t) I don't understand why (uint64_t) casting would be necessary. Maybe I'm missing something...

libsledge/src/memory_instructions.c
```diff
@@ -12,55 +12,55 @@ extern thread_local struct sledge_abi__wasm_module_instance sledge_abi__current_
...
 }
 
 // These functions are equivalent to those in wasm_memory.h, but they minimize pointer dereferencing
 INLINE float
 get_f32(uint32_t offset)
 {
 	assert(sledge_abi__current_wasm_module_instance.memory.buffer != NULL);
-	assert(offset + sizeof(float) <= sledge_abi__current_wasm_module_instance.memory.size);
+	assert((uint64_t)offset + sizeof(float) <= sledge_abi__current_wasm_module_instance.memory.size);
```
emil916, line 23: same as above. why cast here?

libsledge/src/table_instructions.c
```diff
@@ -17,9 +17,7 @@ wasm_table_get(struct sledge_abi__wasm_table *wasm_table, uint32_t idx, uint32_t
 
 	struct sledge_abi__wasm_table_entry f = wasm_table->buffer[idx];
 
-	/* Wasmception-based modules trigger function type mismatches for an unknown reason. 
-	 * This should be reenabled when WASI is added */
-	// assert(f.type_id != type_id);
+	assert(f.type_id == type_id);
```
emil916, line 20: Yay! 👍

runtime/Makefile
```diff
@@ -42,6 +42,12 @@ BINARY_NAME=sledgert
 
 # Debugging Flags
 
+# Enables logs of WASI syscalls
+# CFLAGS += -DLOG_WASI
+
+# Writes sandbox stderr to host stderr
```
emil916, line 48: Where is stderr written to if the flag is switched off? Maybe include in the comment.

runtime/include/scheduler.h
```diff
@@ -254,6 +254,10 @@ scheduler_preemptive_sched(ucontext_t *interrupted_context)
 	/* Preempt executing sandbox */
 	scheduler_log_sandbox_switch(interrupted_sandbox, next);
 	sandbox_preempt(interrupted_sandbox);
+
+	// Write back global at idx 0
+	sledge_abi__wasm_globals_set_i64(0, sledge_abi__current_wasm_module_instance.abi.wasmg_0, true);
```
emil916, line 259: I am having hard time to recall why this needed to be done during the preemptive_sched and not the cooperative_sched?

runtime/src/sledge_abi.c
```diff
@@ -7,7 +9,11 @@ sledge_abi__wasm_memory_expand(struct sledge_abi__wasm_memory *wasm_memory, uint
 {
 	int32_t old_page_count = wasm_memory->size / WASM_PAGE_SIZE;
 
+	struct sandbox *sandbox = current_sandbox_get();
+	sandbox_syscall(sandbox);
```
emil916, line 13: Good catch!

emil916 (approved):
Just a few comments or refactor requests. LGTM!

.gitmodules
bushidocodes: Hopefully should be able to get merged. It's just a bit slower to merge stuff over there generally.

applications/Makefile
bushidocodes: It's actually unused. I seem to recall that I can't use a variable withing a Make rule's name. Pretty goofy TBH!

applications/Makefile
bushidocodes: It was down in the rule itself for some reason. Moved back up here.

libsledge/src/memory_instructions.c
bushidocodes: Yes, it's a uint64_t. UNIT32_MAX is the largest size for wasm32, but it's one too large to fit in a uint32_t.  I've been bitten by type issues in this layer before, so I ended up making everything pretty explicit.

libsledge/src/memory_instructions.c
bushidocodes: It does look unnecessary I think. I'm just not sure if type mismatches might affect stuff like turning this division into a >> right shift since the denominator is a power of two. Mostly just me being paranoid.

runtime/Makefile
bushidocodes: Good idea!

runtime/src/current_sandbox.c
```diff
@@ -10,12 +10,16 @@
...
-#include "sandbox_setup_arguments.h"
 #include "scheduler.h"
 #include "software_interrupt.h"
+#include "wasi_impl.h"
 
 thread_local struct sandbox *worker_thread_current_sandbox = NULL;
 
+// TODO: Propagate arguments from *.json spec file
+const int   dummy_argc   = 1;
+const char *dummy_argv[] = { "Test" };
```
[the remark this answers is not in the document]
bushidocodes: That's right! For wasm, it's usually just the executable name because the path info is privileged. We're not respecting that currently, but I don't think it's a big deal. Each sandbox would just think it's named "Test" as is.

runtime/include/scheduler.h
bushidocodes: Really good catch! I think you're right. This would be a problem when a sandbox blocks. Making the change and will run some additional tests shortly.

bushidocodes: Also not really sure when I'm calling the "ABI" here since we're inside of the runtime already. I'm think these should probably be `wasm_globals_set_i64` instead

Issue #242, opened 2021-06-03
Possible permission issues doing fresh build of dev container


bushidocodes:
Xiaosu Lyu reported permissions issues performing a clean build of the dev container.

I was not able to replicate the issue on a fresh install on my machine, so requires further root cause analysis.

![image](https://user-images.githubusercontent.com/8700322/120691344-54081b80-c474-11eb-9395-82090f51213b.png)

bushidocodes:
Closing this as obsolete — the specific failure reported here can no longer occur, because the build step it failed on has been removed.

### What this issue actually was

The screenshot shows the failure was a **filesystem write-permission error** during the in-container build (not a Dockerfile step):

```
cd ./awsm/wasmception && wget .../wasmception-linux-x86_64-0.2.0.tar.gz -O wasmception.tar.gz && tar xvfz ...
wasmception.tar.gz: Permission denied
make: *** [Makefile:12: build] Error 1
```

i.e. the build couldn't write `wasmception.tar.gz` into `awsm/wasmception/` on the bind-mounted host tree, because that directory contained root-owned files left by a prior build.

### Why it's no longer reproducible

That whole "download the wasmception tarball with `wget`" mechanism was reworked years ago and no longer exists in the tree:

- `4ffc0c7` download wasmception → `def9741` install wasmception after bind mount → `5d19891` libsledge pattern for wasmception (#299)

The current `awsm` build is `cargo build --release` plus the libsledge pattern; there is no `wasmception.tar.gz` download anywhere (`grep -r wasmception.tar.gz` comes back empty). The dev container itself was also restructured since this report (non-root build user in `c01ca73`, and the `/opt/sledge` → `/sledge` layout in `7cacac6`/#267), and `fix_root.sh` exists specifically to clean up root-owned files in the bind mount that caused this class of error.

### Note

To be clear, this is **not** being closed by #388 — that PR repairs other Docker/native build issues (git `safe.directory`, `hey`, Rust env, the `wasm_apps` symlink) and does not touch this code path. This is simply stale: the failing step is gone.

If a permission error on the bind-mounted tree resurfaces with the current build, please open a fresh issue with up-to-date logs rather than reusing this one.

PR #279, opened 2021-10-12
A simple solution to fix compiling permission error of issue #242

lyuxiaosu:
This is a simple solution to fix the compiling permission error related to issue #242 . The partial reason to cause #242 is the container tries to write some files to the host disk during the compilation but it does not have the root permission. This solution is to give the container a root permission to write files to the host disk. Since the solution doesn't consider the security issue if we give it the root permission, it is a simple solution, not a perfect solution. But it could help anyone who are blocked by this compiling  error.


bushidocodes (approved):
Approved as a stopgap until we revisit this.

PR #284, opened 2021-11-19
feat: replay deferred sigalrms

bushidocodes:


runtime/include/sandbox_set_as_running_user.h
```diff
@@ -42,9 +42,11 @@ sandbox_set_as_running_user(struct sandbox *sandbox, sandbox_state_t last_state)
 	runtime_sandbox_total_increment(SANDBOX_RUNNING_USER);
 	runtime_sandbox_total_decrement(last_state);
 
-	/* WARNING: This state change needs to be at the end of this transition because all code below this assignment
-	 * is preemptable */
+	/* WARNING: All code below this assignment is preemptable */
 	sandbox->state = SANDBOX_RUNNING_USER;
+
+	/* Now that we are preemptable, we can replay deferred sigalrms */
+	if (software_interrupt_deferred_sigalrm > 0) raise(SIGALRM);
```
gparmer, line 49: Don't you need to modify `software_interrupt_deferred_sigalrm` to be one fewer for each signal you raise? Atomically?

Do we need to worry about if SIGUSR1 is the signal to send instead?

runtime/include/sandbox_set_as_running_user.h
```diff
@@ -42,9 +42,11 @@ sandbox_set_as_running_user(struct sandbox *sandbox, sandbox_state_t last_state)
 	runtime_sandbox_total_increment(SANDBOX_RUNNING_USER);
 	runtime_sandbox_total_decrement(last_state);
 
-	/* WARNING: This state change needs to be at the end of this transition because all code below this assignment
-	 * is preemptable */
+	/* WARNING: All code below this assignment is preemptable */
 	sandbox->state = SANDBOX_RUNNING_USER;
+
+	/* Now that we are preemptable, we can replay deferred sigalrms */
+	software_interrupt_deferred_sigalrm_replay();
```
gparmer, line 49: You'll need a `volatile asm("")` above this line, I believe. This is a compiler barrier that disallows the compiler to reorder the setting of state after this function.

runtime/src/software_interrupt.c
```diff
@@ -78,16 +37,22 @@ extern pthread_t *runtime_worker_threads;
...
  * This function broadcasts the sigalarm signal to all other worker threads
  */
 static inline void
-sigalrm_propagate_workers(siginfo_t *signal_info)
+propagate_sigalrm(siginfo_t *signal_info)
 {
 	/* Signal was sent directly by the kernel, so forward to other threads */
 	if (signal_info->si_code == SI_KERNEL) {
-		atomic_fetch_add(&software_interrupt_SIGALRM_kernel_count, 1);
+		software_interrupt_counts_sigalrm_kernel_increment();
```
gparmer, line 55: Can take this out of the conditionals as it is the first op in the `if` and the `else`.

gparmer (commented):
A little more complex than I was expecting, but you refactored a lot of the counting infrastructure so I think I'm just seeing the complexity of that.

Looks good to me, but I think that I understand about 1/8 of the implications ;-)

runtime/include/sandbox_set_as_running_user.h
bushidocodes: I defined a Linux-like macro for this and added the barrier to both sides of the state transitions that enable / disable preemption. Might be overkill, but I assume code migration across the state assignment from either direction is dangerous.

runtime/src/software_interrupt.c
bushidocodes: They're actually different functions. One is `*_sigalrm_kernel_increment` and the other is `*_sigalrm_thread_increment`. It help track the amount of times a worker receives the original SIGALRM from the kernel versus a propagated SIGALRM from another thread. The kernel count generally shows that the lowest numbered worker generally gets most of the SIGALRMs form the kernel. The user count shows how efficient the "smart triage" logic is. It seems to reduce interrupts by ~95% for all other core.

runtime/include/sandbox_set_as_running_user.h
bushidocodes: I don't really see value in queuing the SIGALRMs. I assume that if `software_interrupt_deferred_sigalrm` is 1..INF, we should only trigger at most a single replay. I figure the main reason that this isn't a boolean is so we can track the max number of deferred sigalrms for debugging. Still need to think through the SIGUSR1 path a bit.

bushidocodes: I went through the SIGUSR1 path when a cooperative sched executes the scheduler outside of a signal, and the scheduler wants to execute a sandbox in the PREEMPTED state, and it seems to be fine. The worker updates its current_sandbox to the preempted sandbox, but the sandbox doesn't transition to RUNNING_USER until inside of the SIGUSR1 signal handler (when signals are masked), so it doesn't seem like there is any point where something is unintentionally preemptable.

runtime/include/sandbox_set_as_running_user.h
gparmer: Both sides isn't necessary. When disabling preemptions, you want to make sure that accesses in the runtime aren't reordered before the disable. When enabling preemptions, you want to make sure that previous runtime code is not reordered after the enable. I'd rather encourage people not overusing concurrency primitives, here.

runtime/include/sandbox_set_as_running_user.h
gparmer: I cannot follow this logic, and it seems more complicated than resetting the value before triggering deferred signals. Complexity is the enemy for this type of code.

If multiple sigalarms are queued, isn't that a bug? "Tracking for debugging" only makes sense if it shouldn't be a panicable behavior.

bushidocodes: An contrived example of "multiple queued sigalrms" might be a long-running WASI syscall that takes let's say 20ms running on the worker that is receiving the kernel SIGALRMs. The syscall is interrupted and receives at least three SIGALRMs, each of which increments `software_interrupt_deferred_sigalrm` by 1. When preemption is finally restored, the value of `software_interrupt_deferred_sigalrm` is 3. When the deferred sigalrms are replayed, `software_interrupt_deferred_sigalrm` is set to 0 and a single SIGALRM is raised. If we treat "multiple queued sigalrms" as a panicable behavior, that seems to imply that we need to be able to make hard guarantees that a worker will never execute more than 10ms (2x quantum) of control flow in a non-preemptable state.

gparmer: You're absolutely correct. Thanks for walking me through it!

bushidocodes: For the other point of "why doesn't replay clear the deferred sigalrms?," basically I clear the deferred sigalrms in two places currently: `scheduler_cooperative_sched` and `scheduler_preemptive_sched`. The reason for this is that our runtime might be calling the scheduler cooperatively while in a non-preemptive state. This is what happens when a worker is idling in what we call the `base_context`. In this state, a worker's `current_sandbox` is NULL, which means that the worker is non-preemptable. However, the base context just repeatedly calls `scheduler_cooperative_sched` over and over. This means that the scheduling logic is getting run. However, the idle loop does still respond to SIGALRMs, and the handler would still call `defer_sigalrm` when when interrupting the cooperative scheduler. Because of this, all invocations of cooperative scheduler end up zeroing this value back out. This sort of seems like a hack.

After talking this though, I think that the SIGALRM handler should perhaps only call  `defer_sigalrm` when current_sandbox is not NULL and in a non-preemptive state. If `current_sandbox` is NULL, that signifies that the worker thread is already in the scheduler in some form, so we can ignore the SIGALRM. Once we do that, we can running `software_interrupt_deferred_sigalrm_clear` in `scheduler_cooperative_sched`. That allows us to move the logic of  `software_interrupt_deferred_sigalrm_clear` into `software_interrupt_deferred_sigalrm_replay`.

Assuming that deferred_sigalrm stays a counter, rather than a bool. That ends up looking something like:

```c
static inline void
software_interrupt_counts_deferred_sigalrm_max_update(int deferred_sigalrm_count)
{
#ifdef LOG_SOFTWARE_INTERRUPT_COUNTS
	if (unlikely(deferred_sigalrm_count > software_interrupt_counts_deferred_sigalrm_max[worker_thread_idx])) {
		software_interrupt_counts_deferred_sigalrm_max[worker_thread_idx] = deferred_sigalrm_count;
	}
#endif
}

static inline void
software_interrupt_deferred_sigalrm_clear()
{
	software_interrupt_counts_deferred_sigalrm_max_update(deferred_sigalrm);
	atomic_store(&deferred_sigalrm, 0);
}

static inline void
software_interrupt_deferred_sigalrm_replay()
{
	if (deferred_sigalrm > 0) {
		software_interrupt_deferred_sigalrm_clear();
		software_interrupt_counts_deferred_sigalrm_replay_increment();
		raise(SIGALRM);
	}
}
```

There is a bunch of overhead due to the debugging stuff. If deferred sigalrm was treated as a bool, we could for sure just have a compare and swap that conditionally triggers a RAISE. Maybe we could do that if `LOG_SOFTWARE_INTERRUPT_COUNTS` is not set since we wouldn't be using the count in that case.

runtime/include/runtime.h
```diff
@@ -8,6 +8,9 @@
 #include "likely.h"
 #include "types.h"
 
+/* Memory barrier to prevent unexpected code migration due to compiler optimization */
+#define barrier() __asm__ __volatile__("" ::: "memory")
```
emil916, line 12: I don't get it. Unexpected code migration from what to what?

runtime/include/sandbox_set_as_running_user.h
emil916: Ok nevermind. Got my answer to the question above.

runtime/Makefile
```diff
@@ -119,7 +119,7 @@ all: runtime
 bin/sledgert: ${CFILES}
 	@echo "Compiling runtime"
 	@mkdir -p bin/
-	${CC} ${INCLUDES} ${CFLAGS} ${LDFLAGS} ${JSMNCFLAGS} -L/usr/lib/ $^ -o bin/sledgert
+	${CC} ${CFLAGS} ${LDFLAGS} ${JSMNCFLAGS} ${INCLUDES}  -L/usr/lib/ $^ -o bin/sledgert
```
emil916, line 122: What difference does this make?

emil916 (commented):
This all makes sense. Just so I understand it correctly, a signal can be deffered only during the cooperative_scheduling, correct? Cuz, if in preemptive_scheduling then it'll consider as a "nested signal", which are not allowed anwyays?

runtime/include/runtime.h
bushidocodes: Sorry for the delay here! Basically, an optimizing compiler will reorder instructions however it sees fit. If there is a variable protecting a critical section, this means that code may move out of or into the critical section, which could cause bugs. Inline assembly acts as a barrier that no assembly instructions can be reordered across. For us, we use this to ensure that code is either intentionally preemptable or non-preemptable.

runtime/Makefile
bushidocodes: I think that I was trying to use one of the debugging flags, and it seemed to not be working as expected. I think I had a hypothesis that the flags we are adding to CFLAGS were not evaluating as expected in our header files. I'm not actually sure if this made an actual difference at this point.

bushidocodes: Ran a small experiment to validate. The order shouldn't matter. I'll revert the change on this line.

bushidocodes:
@emil916 - When a SIGALRM fires, a worker can be in one of four states:

1. "Running a signal handler" - We mask signals when we are executing a signal handler, which results in signals being ignored. A kernel signal should get delivered to another unmasked worker, so propagation still occurs. 
2. "Running the Cooperative Scheduler" - This is signified by the thread local `current_sandbox` being set to NULL. We propagate the signal and return immediately because we know we're already in the scheduler. We have no sandboxes to interrupt, so no sandbox state transitions occur. 
3. "Running a Sandbox in a state other than `SANDBOX_RUNNING_USER`" - We call `sandbox_interrupt` on `current_sandbox`, propagate the sigalrms to the other workers, defer the sigalrm locally, and then return. The `SANDBOX_INTERRUPTED` timekeeping data is increased to account for the time needed to propagate the sigalrms.
4. "Running a Sandbox in the `SANDBOX_RUNNING_USER` state` - We call `sandbox_interrupt` on `current_sandbox`, propagate the sigalrms to the other workers, and then actually enter the scheduler via `scheduler_preemptive_sched`. The interrupted sandbox may either be preempted or return to depending on the scheduler. The `SANDBOX_INTERRUPTED` timekeeping data is increased to account for the time needed to propagate the sigalrms, run epoll, query the scheduler data structure, and (potentially) allocate and initialize a sandbox.

So a signal is NOT deferred when in cooperative scheduling because we are already executing the scheduling logic, so deferring a SIGALRM would be functionally redundant. We only run the preemptive scheduler in a signal handler, which masks signals and prevents nesting. This means that we only defer sigalrms when we (A) not in the scheduler (preemptive or cooperative) and (B) running a sandbox that is not in "userspace." This can be WASI calls, per-sandbox initialization, teardown, etc.

bushidocodes:
@emil916 - If this makes sense, perhaps I can add this as a block comment in scheduler.h or something? Let me know!

runtime/src/software_interrupt.c
```diff
@@ -144,55 +99,65 @@ software_interrupt_handle_signals(int signal_type, siginfo_t *signal_info, void
...
+			propagate_sigalrm(signal_info);
+			break;
 		}
 
-		scheduler_preemptive_sched(interrupted_context);
+		sandbox_state_t interrupted_state = current_sandbox->state;
+		sandbox_interrupt(current_sandbox);
+		propagate_sigalrm(signal_info);
+
+		if (interrupted_state == SANDBOX_RUNNING_USER) {
```
emil916, line 119: I liked the `!sandbox_is_preemptable(current_sandbox)` more.

emil916:
> This means that we only defer sigalrms when we (A) not in the scheduler (preemptive or cooperative) and (B) running a sandbox that is not in "userspace." This can be WASI calls, per-sandbox initialization, teardown, etc.

Aren't pre-sandbox initialization and sandbox teardown technically part of the scheduling process (preemptive or cooperative)?

emil916 (approved):
Great control flow! Though, this will make my work a bit more complicated, but we'll think about that later :)

PR #287, opened 2021-12-05
feat: WIP restage of memory allocation features

bushidocodes:
This PR contains a number of refactors:
- Eliminate the "Sandbox Request" and instead modify all global scheduler logic to allocate a "Sandbox" directly.
- Begin to decompose parts of a "Sandbox" into objects with OOP-style alloc, init, free methods and associated logic
- Cleanup the HTTP Request and Response Logic to use discrete buffers that do not make assumptions about layout.


runtime/src/sandbox_state.c
```diff
@@ -23,31 +23,6 @@ const char *sandbox_state_labels[SANDBOX_STATE_COUNT] = {
...
 #endif
-
-/*
- * Function intended to be interactively run in a debugger to look at sandbox totals
- * via `call runtime_log_sandbox_states()`
- */
-void
-runtime_log_sandbox_states()
-{
-#ifdef LOG_SANDBOX_COUNT
```
bushidocodes, line 37: I believe that I removed this function because it was trivial to log this stuff manually with gdb commands.

runtime/src/libc/syscall.c
```diff
@@ -318,25 +320,20 @@ wasm_mremap(int32_t offset, int32_t old_size, int32_t new_size, int32_t flags)
...
+	int32_t new_offset = current_wasm_module_instance.memory->size;
+	wasm_memory_expand(current_wasm_module_instance.memory, new_size);
 
 	// Get pointer of old offset and pointer of new offset
-	char *linear_mem = local_sandbox_context_cache.memory.start;
-	char *src        = &linear_mem[offset];
-	char *dest       = &linear_mem[new_offset];
+	uint8_t *linear_mem = current_wasm_module_instance.memory->data;
+	uint8_t *src        = &linear_mem[offset];
+	uint8_t *dest       = &linear_mem[new_offset];
```
bushidocodes, line 336: Generally, my tendency is to use `uint8_t` in place of char for abstract bytes at this point and to leave char for actual strings.

runtime/Makefile
```diff
@@ -73,10 +73,9 @@ BINARY_NAME=sledgert
 # To log, run `call http_total_log()` while in GDB
 # CFLAGS += -DLOG_TOTAL_REQS_RESPS
 
-# This flag logs the total number of sandboxes in the various states
+# This flag tracks the total number of sandboxes in the various states
 # It is useful to debug if sandboxes are "getting caught" in a particular state
-# To log, run `call runtime_log_sandbox_states()` while in GDB
-# CFLAGS += -DLOG_SANDBOX_COUNT
+# CFLAGS += -DSANDBOX_STATE_TOTALS
```
bushidocodes, line 78: I changed this flag to enable or disable per state atomic counts. I eliminated the bespoke function to print this as it's trivial to pretty print the structure directly in gdb. IMO, I think our debugging code become substantially more maintainable when we focus on accumulating debugging state that we can inspect inside gdb. This sort of presumes that all contributors get comfortable with gdb.

runtime/compiletime/memory_instructions.c
```diff
@@ -1,78 +1,48 @@
...
 
-#include "types.h"
+#include "wasm_module_instance.h"
 
-uint32_t
+INLINE uint32_t
 instruction_memory_size()
 {
-	return local_sandbox_context_cache.memory.size / WASM_PAGE_SIZE;
+	return wasm_memory_get_page_count(current_wasm_module_instance.memory);
```
bushidocodes, line 8: `current_wasm_module_instance` is how we associate calls to the backing compiletime functions for wasm32 instructions with the running sandbox. I renamed this as "current_wasm_module_instance" since the only state that a wasm32 instruction should ever touch is the portion of the sandbox that aligns to a "wasm_module_instance." In the future, I imagine a sandbox struct will be refactored to contain a wasm_module_instance struct.

runtime/compiletime/memory_instructions.c
```diff
@@ -1,78 +1,48 @@
...
 INLINE float
 get_f32(uint32_t offset)
 {
-	assert(offset + sizeof(float) <= local_sandbox_context_cache.memory.size);
-
-	char *mem_as_chars = (char *)local_sandbox_context_cache.memory.start;
-	void *address      = &mem_as_chars[offset];
-
-	return *(float *)address;
+	return wasm_memory_get_f32(current_wasm_module_instance.memory, offset);
```
bushidocodes, line 15: These functions are all symbols that back invocations that are code-gened by aWsm. This means for example that if we wanted to namespace `get_f32`, we would have to change aWsm, SLEdge, all other runtimes, and recompile all modules.

runtime/include/client_socket.h
```diff
@@ -32,53 +33,55 @@ client_socket_close(int client_socket, struct sockaddr *client_address)
...
-
-			debuglog("Error with %s\n", strerror(errno));
-
-			goto send_err;
+			if (errno == EAGAIN) {
+				if (on_eagain == NULL) {
+					rc = -1;
+					goto done;
+				}
+				on_eagain();
```
bushidocodes, line 60: I added the `on_eagain` callback here because I didn't want to intermingle the scheduler code here. The scheduler touches so much of the codebase that it makes our translation unit graph a tangle. I view callbacks as a reasonable want to break this up.

runtime/include/wasm_memory.h
```diff
@@ -0,0 +1,302 @@
...
+ * @param offset an offset into the WebAssembly linear memory
+ * @param size the maximum expected length in characters
+ * @return pointer to the string or NULL if max_length is reached without finding null-terminator
+ */
+static inline char *
+wasm_memory_get_string(struct wasm_memory *self, uint32_t offset, uint32_t size)
+{
+	assert(offset + (sizeof(char) * size) <= self->size);
+
+	for (uint32_t i = 0; i < size; i++) {
```
gparmer, line 207: isn't this effectively `strnlen`? glibc will have very fast versions by default.

runtime/include/wasm_memory.h
```diff
@@ -0,0 +1,302 @@
...
+/**
+ * Translates WASM offsets into runtime VM pointers
+ * @param offset an offset into the WebAssembly linear memory
+ * @param bounds_check the size of the thing we are pointing to
+ * @return void pointer to something in WebAssembly linear memory
+ */
+static inline void *
+wasm_memory_get_ptr_void(struct wasm_memory *self, uint32_t offset, uint32_t size)
+{
+	assert(offset + size <= self->size);
```
gparmer, line 102: We obviously want to change these from `assert` to some form of sandbox error that sledge deals with. Pulling them out into a macro here (to maintain the sloc) might be good? To consider for the futurel.

runtime/include/wasm_memory.h
```diff
@@ -0,0 +1,302 @@
...
+{
+	return self->size;
+}
+
+static inline void
+wasm_memory_initialize_region(struct wasm_memory *self, uint32_t offset, uint32_t region_size, uint8_t region[])
+{
+	assert((size_t)offset + region_size <= self->size);
+	memcpy(&self->data[offset], region, region_size);
+}
```
gparmer, line 302: I find all of this a little odd. You're just moving stuff that was already in a function per operation into effectively another set of functions per op by adding a level of indirection. I don't see what you buy by doing this, but am not opposed to it.

runtime/include/wasm_stack.h
```diff
@@ -0,0 +1,68 @@
...
+static inline void
+wasm_stack_free(struct wasm_stack *stack)
+{
+	assert(stack != NULL);
+	assert(stack->buffer != NULL);
+	/* The stack start is the bottom of the usable stack, but we allocated a guard page below this */
+	int rc        = munmap(stack->buffer, stack->capacity + PAGE_SIZE);
+	stack->buffer = NULL;
+	if (unlikely(rc == -1)) perror("munmap");
+}
```
gparmer, line 68: You have any idea what happens if we set up a signal for SIG_TRAP, then trap out by hitting the guard page here? Normally the signal would execute on the stack....but....

runtime/include/wasm_table.h
```diff
@@ -0,0 +1,66 @@
...
+}
+
+static inline void *
+wasm_table_get(struct wasm_table *self, uint32_t idx, uint32_t type_id)
+{
+	assert(self != NULL);
+	assert(idx < self->capacity);
+
+	struct wasm_table_entry f = self->data[idx];
+	// FIXME: Commented out function type check because of gocr
```
gparmer, line 47: Yikes! Maybe keep it by default and have a #define or something?

runtime/include/http.h
```diff
@@ -30,46 +33,55 @@
...
-http_response_200_size(char *content_type, char *content_length)
+static inline int
+http_header_200_write(int fd, const char *content_type, size_t content_length)
 {
-	size_t size = 0;
-	size += strlen(HTTP_RESPONSE_200_TEMPLATE) - HTTP_RESPONSE_200_TEMPLATE_FORMAT_SPECIFIER_LENGTH;
-	size += strlen(content_type);
-	size += strlen(content_length);
-	return size;
+	return dprintf(fd, HTTP_RESPONSE_200_TEMPLATE, content_type, content_length);
```
bushidocodes, line 46: I was allocating a temporary in-memory buffer to sprintf to. This seemed silly when I realized I could dprintf directly instead. I forget about a bunch of these I/O functions, so it didn't occur to me until I skimmed APUE one night recently.

runtime/src/libc/syscall.c
```diff
@@ -318,25 +320,20 @@ wasm_mremap(int32_t offset, int32_t old_size, int32_t new_size, int32_t flags)
...
-	int32_t new_offset = local_sandbox_context_cache.memory.size;
-	for (int i = 0; i < pages_to_allocate; i++) expand_memory();
+	int32_t new_offset = current_wasm_module_instance.memory->size;
+	wasm_memory_expand(current_wasm_module_instance.memory, new_size);
 
 	// Get pointer of old offset and pointer of new offset
-	char *linear_mem = local_sandbox_context_cache.memory.start;
-	char *src        = &linear_mem[offset];
-	char *dest       = &linear_mem[new_offset];
+	uint8_t *linear_mem = current_wasm_module_instance.memory->data;
```
gparmer, line 334: Can't tell if the additional level of indirection is on the fastpath (memory access) or not. I'll assume not, unless told otherwise ;-)

runtime/src/sandbox.c
```diff
@@ -94,58 +55,70 @@ sandbox_allocate_stack(struct sandbox *sandbox)
...
- * @param sandbox_request request being allocated
- * @returns sandbox * on success, NULL on error
+ * Allocates HTTP buffers and performs our approximation of "WebAssembly instantiation"
+ * @param sandbox
+ * @returns 0 on success, -1 on error
  */
-struct sandbox *
-sandbox_allocate(struct sandbox_request *sandbox_request)
+int
+sandbox_prepare_execution_environemnt(struct sandbox *sandbox)
```
gparmer, line 105: spelling. Isn't vscode smart enough to help with this ;-)

gparmer (approved):
Don't see any blockers here. Great job. Just some comments throughout.

runtime/include/sandbox_setup_arguments.h
```diff
@@ -15,12 +17,13 @@ sandbox_setup_arguments(struct sandbox *sandbox)
...
-	sandbox->arguments_offset = local_sandbox_context_cache.memory.size;
-	assert(local_sandbox_context_cache.memory.start == sandbox->memory.start);
-	expand_memory();
 
-	int32_t string_off = sandbox->arguments_offset;
+	/* Copy arguments into linear memory. It seems like malloc would clobber this, but I think this goes away in
+	 * WASI, so not worth fixing*/
+
+	sandbox->arguments_offset = wasm_memory_get_size(sandbox->memory);
+	int rc                    = wasm_memory_expand(sandbox->memory, WASM_PAGE_SIZE);
```
bushidocodes, line 25: I do not have a good understanding of wasmception libc init logic. It seemed like reading the initial and max page size from the module caused segfaults. I believe I padded this by a page to get this working. This is likely a hack, but I didn't see much sense in investing in learning wasmception internals at this point.

runtime/include/sandbox_state_history.h
```diff
@@ -3,12 +3,30 @@
...
 
+#ifdef LOG_STATE_CHANGES
+#define SANDBOX_STATE_HISTORY_CAPACITY 100
+#else
+#define SANDBOX_STATE_HISTORY_CAPACITY 0
+#endif
+
+struct sandbox_state_history {
+	uint16_t        size;
+	sandbox_state_t buffer[SANDBOX_STATE_HISTORY_CAPACITY];
```
bushidocodes, line 14: Brute force approach to saving this state. I've never seen a sandbox have enough transitions to exhaust this buffer.

runtime/src/libc/syscall.c
gparmer: I don't care about the data-type. I was commenting about the `->`.

runtime/include/vec_u8.h
```diff
@@ -0,0 +1,37 @@
+#pragma once
+
+#include <stdint.h>
+#include <stdlib.h>
+
+struct vec_u8 {
```
bushidocodes, line 6: The intention here is that this header will be gradually enhanced to add typical push, pop APIs that realloc the buffer as required. I imagine this will probably end up wrapped as a macro to enable generics, but I've also considered keeping the buffer uint8_t[] and instead add a type tag that gets populated from a C11 _Generic macro.

runtime/include/wasm_memory.h
```diff
@@ -0,0 +1,302 @@
...
+
+struct wasm_memory {
+	size_t  size;     /* Initial Size in bytes */
+	size_t  capacity; /* Size backed by actual pages */
+	size_t  max;      /* Soft cap in bytes. Defaults to 4GB */
+	uint8_t data[];
+};
+
+static inline struct wasm_memory *
+wasm_memory_allocate(size_t initial, size_t max)
```
bushidocodes, line 23: I've been inconsistent with the naming conventions here. I believe the follow up memory pool renames this to `wasm_memory_new`. I'm essentially following C++ semantics. "allocate" means allocate the OOP type on the heap. "init" means initialize a preallocated object, including allocating member buffers as needed. "reinit" means reinitialize members and wipe member buffers for reuse. "deinit" means cleanup and free member buffers. "free" means free the base struct from the heap. "new' means alloc + init. "delete" means deinit + free.

bushidocodes: I'm also generally inconsistent betwen alloc versus allocate. init versus initialize. I also sometimes name the first argument the same as the struct name and othertimes I name it self (as in Rust). Just some discussion points for things that might be work standardizing on and cleaning up in a follow-up refactor.

runtime/src/global_request_scheduler_deque.c
```diff
@@ -1,46 +1,48 @@
...
- * Pushes a sandbox request to the global deque
- * @param sandbox_request
+ * Pushes a sandbox to the global deque
+ * @param sandbox_raw
  * @returns pointer to request if added. NULL otherwise
  */
-static struct sandbox_request *
-global_request_scheduler_deque_add(void *sandbox_request_raw)
+static struct sandbox *
+global_request_scheduler_deque_add(void *sandbox_raw)
```
bushidocodes, line 18: I'm unclear why this type is a `void *`.

runtime/src/global_request_scheduler_deque.c
```diff
@@ -1,46 +1,48 @@
 #include "global_request_scheduler.h"
+#include "global_request_scheduler_deque.h"
 #include "runtime.h"
 
+#define GLOBAL_REQUEST_SCHEDULER_DEQUE_CAPACITY (1 << 19)
+
 static struct deque_sandbox *global_request_scheduler_deque;
 
 /* TODO: Should this be used???  */
 static pthread_mutex_t global_request_scheduler_deque_mutex = PTHREAD_MUTEX_INITIALIZER;
```
bushidocodes, line 10: Check if this is zombie code.

runtime/src/module.c
```diff
@@ -182,29 +183,34 @@ module_new(char *name, char *path, uint32_t stack_size, uint32_t max_memory, uin
 	admissions_info_initialize(&module->admissions_info, admissions_percentile, expected_execution,
 	                           module->relative_deadline);
 
+	/* WebAssembly Indirect Table */
+	/* TODO: Should this be part of the module or per-sandbox? */
+	/* TODO: How should this table be sized? */
```
bushidocodes, line 188: These TODOs seem to relate to discussions around the "store" and module instantiation, so likely worth deferring.

runtime/src/sandbox.c
bushidocodes, line 105: The semantics here are a bit of a mess. The "execution environment" is a webassembly module instance. It doesn't totally line up because this also allocates the http buffers.

bushidocodes: It should be!

runtime/include/types.h
```diff
@@ -20,18 +18,10 @@
...
-	void *   func_pointer;
-};
-
-/* Cache of Frequently Accessed Members used to avoid pointer chasing */
-struct sandbox_context_cache {
-	struct wasm_memory           memory;
-	struct indirect_table_entry *module_indirect_table;
-};
+#ifndef unlikely
+#define unlikely(x) __builtin_expect(!!(x), 0)
```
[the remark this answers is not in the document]
bushidocodes: Yeah, C and Unix have sort of beat a lot of dogma out of me at this point TBH. I'm not quite to CREAT, but I'm close.

runtime/include/wasm_memory.h
bushidocodes: I would say the main reason is to achieve a layer of abstraction from the ABI symbols that aWsm requires. We can internally organize the state and behavior for wasm components in a cleaner way.

runtime/include/wasm_stack.h
bushidocodes: @gparmer - I would assume the signal handler would panic the process or something. I seem to recall that you can consider signal handlers to use a dedicated stack. Perhaps we should be doing that so a module stack overflow is recoverable?

runtime/include/wasm_table.h
bushidocodes: This ugly baby has been around for a while. I'll flag an issue for this and revisit this after we merge.

bushidocodes: Opened #289

runtime/src/libc/syscall.c
bushidocodes: I'll revisit this. I think I can restructure things to flatten them out as long as I drop flexible array members entirely and replace them with pointers.

runtime/include/wasm_stack.h
bushidocodes: Captured as #290

runtime/src/global_request_scheduler_deque.c
bushidocodes: Ambiguous. Opened #291 as a follow-on task. Non-blocking because this did not change in this PR.

runtime/src/global_request_scheduler_deque.c
bushidocodes: #292

PR #288, opened 2021-12-08
Reduce Allocations via Object Pools 

bushidocodes:


runtime/src/sandbox.c
```diff
@@ -218,7 +213,7 @@ sandbox_free(struct sandbox *sandbox)
 
 	/* Free Sandbox Struct and HTTP Request and Response Buffers */
 
-	if (likely(sandbox->stack.buffer != NULL)) sandbox_free_stack(sandbox);
+	if (likely(sandbox->stack->buffer != NULL)) sandbox_free_stack(sandbox);
```
bushidocodes, line 216: Adding a note for myself. This should be `if (likely(sandbox->stack->buffer != NULL))` now that the stack is its own heap allocation.

runtime/include/module.h
```diff
@@ -141,12 +158,62 @@ module_release(struct module *module)
...
+static inline struct wasm_memory *
+module_allocate_linear_memory(struct module *module)
+{
+	assert(module != NULL);
+
+
+	size_t initial = (size_t)module->abi.starting_pages * WASM_PAGE_SIZE;
+	size_t max     = (size_t)module->abi.max_pages * WASM_PAGE_SIZE;
+
+	assert(initial <= (size_t)UINT32_MAX + 1);
```
gparmer, line 192: Are we bound to only 64 bit architectures? I don't know if we want to make implicit assumptions about only working on 64 bit architectures, though I doubt we'll ever run on 32.

I think that there is a cleaner way to express this without requiring size_t > int.

gparmer: thinking about this a little more, I noticed the `=` here. Not sure why we would allow it to be over the max.

`assert(initial & (UINT32_MAX - 1) == 0)` ?

assumes UINT32_MAX is a power of 2, and 2s complement. I'm OK with tehse assumptions. Up to you.

runtime/include/module.h
```diff
@@ -141,12 +158,62 @@ module_release(struct module *module)
...
+	assert(max <= (size_t)UINT32_MAX + 1);
+
+	struct wasm_memory *linear_memory = wasm_memory_pool_remove_nolock(&module->pools.memory[worker_thread_idx]);
+	if (linear_memory == NULL) {
+		linear_memory = wasm_memory_new(initial, max);
```
gparmer, line 197: `new` isn't really a thing in C. `_alloc` is usually "allocate and initialize", and `_init` is "initialize a piece of memory". alloc calls init. Init exists as a separate function for stack and global data.

runtime/include/module.h
```diff
@@ -141,12 +158,62 @@ module_release(struct module *module)
...
 }
 
+static inline struct wasm_stack *
+module_allocate_stack(struct module *self)
+{
+	assert(self != NULL);
+
+	struct wasm_stack *stack = wasm_stack_pool_remove_nolock(&self->pools.stack[worker_thread_idx]);
+
+	if (stack == NULL) {
```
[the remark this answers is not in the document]
gparmer: A benefit of treating the pool as a transparent cache is that you don't repeat the allocation code (though you need a function to allocate/dealloc as arguments). Also allows centralization of functionality to manage the size of the pool. However, these benefits are pretty small, so you could do it that way, or have your own explicit wrappers as you're doing here.

runtime/include/pool.h
```diff
@@ -0,0 +1,102 @@
...
+	{                                                                                                         \
+		assert(self != NULL);                                                                             \
+		assert(obj != NULL);                                                                              \
+		assert(self->use_lock);                                                                           \
+                                                                                                                  \
+		LOCK_LOCK(&self->lock);                                                                           \
+		ps_list_head_add_d(&self->list, obj);                                                             \
+		LOCK_UNLOCK(&self->lock);                                                                         \
+		return 0;                                                                                         \
+	}
```
gparmer, line 102: I might be completely missing something here, but I don't see any type specialization that is necessary, nor any fast-paths that require aggressive inlining of the macro arguments. The DTOR could be a function pointer within the structure. Because of this, I don't see why this needs to be a macro rather than a simple generic data-structure. I'm not confident I'm understanding everything here, so I could be wrong.

I don't think it is relevant here, but if you need a lock-free cache for this stuff, https://github.com/gwsystems/ps/blob/master/ps_pgalloc.h

runtime/include/wasm_memory.h
```diff
@@ -42,20 +44,21 @@ wasm_memory_allocate(size_t initial, size_t max)
...
+	ps_list_init_d(self);
 	self->size     = initial;
 	self->capacity = initial;
 	self->max      = max;
 	return self;
 }
 
 static inline void
-wasm_memory_free(struct wasm_memory *self)
+wasm_memory_delete(struct wasm_memory *self)
```
gparmer, line 61: The problem with using C++ naming for a lot of this stuff is that it could lull people into thinking that new and delete function as they do in C++. They do not as both of those operations are quite complex in C++. I'm *guessing* this is why it is rarer to use this naming in C. Conflating ideas can lead to bugs.

runtime/include/wasm_stack.h
```diff
@@ -21,48 +28,93 @@ struct wasm_stack {
  * @returns 0 on success, -1 on error
  */
 static inline int
-wasm_stack_allocate(struct wasm_stack *stack, size_t capacity)
+wasm_stack_init(struct wasm_stack *self, size_t capacity)
```
gparmer, line 31: Something to consider: self vs. stack is losing information, and editors in C aren't going to associate self with the C namespace it is associated with, `wasm_stack`. This means that a reader might need to "scroll up" to recall the object type/namespace being operated on. The upside is that someone writes a lot of self or rust code will know what this means. But no-one who develops on sledge does either (beside you). I'm not suggesting to change this, but I don't see a benefit, and I see a minor downside.

runtime/include/wasm_stack.h
```diff
@@ -21,48 +28,93 @@ struct wasm_stack {
...
-	assert(stack != NULL);
-	assert(stack->buffer != NULL);
+	free(self);
+}
+
+
+static struct wasm_stack *
+wasm_stack_new(size_t capacity)
+{
+	struct wasm_stack *self = wasm_stack_allocate();
```
gparmer, line 77: I think this demonstrates why the common `_alloc` allocates and calls `_init`, and `_init` initializes the structure is useful. I don't think that your `_allocate` is called in places not followed by `_init`, so it might as well do both.

gparmer (approved):
Some notes for consideration. I don't think that the pool needs to be a macro, but could be wrong.

runtime/include/module.h
```diff
@@ -32,12 +40,19 @@
   "MODULE_MAX_PENDING_CLIENT_REQUESTS likely exceeds the value in /proc/sys/net/core/somaxconn and thus may be silently truncated";
 #endif
 
+/* TODO: Dynamically size based on number of threads */
+#define MAX_WORKER_THREADS 64
+
+struct module_pools {
+	struct wasm_memory_pool memory[MAX_WORKER_THREADS];
```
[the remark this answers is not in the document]
bushidocodes: Where might CACHE_LINE come from? Do you have an example of how another project defines this? I refactored as you suggested and added `__attribute__((aligned(8)));` for now.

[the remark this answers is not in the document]
gparmer: We'd have to define it. You actually want `128`, not `8`. Intel uses a double cache-line as a coherency unit (128 bytes).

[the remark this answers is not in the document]
gparmer: Just remembered that you're including ps. 

https://github.com/gwsystems/ps/blob/master/plat/arch/x86/ps_arch_x86_common.h#L31

emil916 (approved):
Looks good! Left some minor comments.

runtime/include/wasm_stack.h
bushidocodes: I guess the main benefit is to communicate that "this is a OOP-style function," but if the self or this aren't meaningful in this way, I might as well roll it back. I've got self in a bunch of places in the repo. I think I'm going to take advantage of the different namespaces for structs and have the "self" argument have the same name as the type. That gives us: 

`static inline int wasm_stack_allocate(struct wasm_stack *wasm_stack, size_t capacity)`

runtime/include/wasm_memory.h
bushidocodes: What else would call "allocate the outermost data structure" and "run the initializer/constructor which allocates members with RAII semantics" (and visa-versa with the deinit/dtor and free)? A lot of the existing APIs were simpler because the existing code made assumptions about being a struct nested as a member in another struct or the buffer being static or other such things?

I think there's a big win to having some opinions about "OOP-like" translation units in our project and the lifecycle of how those work. This is similar to the approach in this book: http://www.r-5.org/files/books/computers/languages/c/mod/David_R_Hanson-C_Interfaces_and_Implementations-EN.pdf

runtime/include/wasm_stack.h
bushidocodes: Reading this after writing my other comment above. I think that this change sounds fine and likely addresses your concerns about using a name that is also a C++ keyword.

bushidocodes: I'll make this change. alloc will wrap init. free will wrap deinit.

PR #299, opened 2022-01-25
feat: libsledge pattern for wasmception

bushidocodes:


bushidocodes:
I noticed this in the logs. I've got a public struct and a private struct that wraps it. I think I need to cast to the public struct here I believe.

```
include/sandbox_setup_arguments.h:31:2: warning: 'memcpy' will always overflow; destination buffer has size 32, but size argument is 48 [-Wfortify-source]
        memcpy(&sledge_abi__current_wasm_module_instance.abi.memory, sandbox->memory, sizeof(struct wasm_memory));
        ^
```

libsledge/src/instantiation.c
```diff
@@ -0,0 +1,60 @@
+#include "sledge_abi.h"
+
+#define EXPORT __attribute__((visibility("default")))
+#define WEAK   __attribute__((weak))
```
emil916, line 4: Maybe a comment for what these are would be good?

libsledge/src/numeric_instructions.c
```diff
@@ -1,13 +1,11 @@
...
 
-#include "types.h"
+#include "sledge_abi.h"
 
-#define CHAR_BIT 8
-
-// TODO: Throughout here we use `assert` for error conditions, which isn't optimal
-// Instead we should use `unlikely` branches to a single trapping function (which should optimize better)
-// The below functions are for implementing WASM instructions
+#define CHAR_BIT    8
```
emil916, line 5: What is the point of keeping this instead of just using `sizeof(uint8)` below?

libsledge/src/table_instructions.c
```diff
@@ -0,0 +1,49 @@
+#include "sledge_abi.h"
+
+#define unlikely(X)         __builtin_expect(!!(X), 0)
+#define INDIRECT_TABLE_SIZE (1 << 10)
+#define INLINE              __attribute__((always_inline))
+#define likely(X)           __builtin_expect(!!(X), 1)
+#define unlikely(X)         __builtin_expect(!!(X), 0)
```
emil916, line 7: Duplicate definition L3 = L7

libsledge/src/table_instructions.c
```diff
@@ -0,0 +1,49 @@
...
+
+static INLINE void *
+wasm_table_get(struct sledge_abi__wasm_table *wasm_table, uint32_t idx, uint32_t type_id)
+{
+	assert(wasm_table != NULL);
+	assert(idx < wasm_table->capacity);
+
+	struct sledge_abi__wasm_table_entry f = wasm_table->buffer[idx];
+
+	// assert(f.type_id != type_id);
```
emil916, line 21: What's this?

libsledge/src/numeric_instructions.c
bushidocodes: No point! I think that seems like a good refactor.

bushidocodes: Actually, I take it back. sizeof returns the size in bytes, but I think the functions below require bits. However, CHAR_BIT is a standard POSIX header apparently, so we can just include limits.h to get this. https://www.gnu.org/software/libc/manual/html_node/Width-of-Type.html

libsledge/src/table_instructions.c
bushidocodes: This is commented out because Wasmception compiled serverless functions trigger type errors. I added a note into the source and expect this to be reenabled in the follow-up WASI PR.

emil916 (approved):
Looks good!

PR #306, opened 2022-02-17
feat: Dynamically size runtime globals

bushidocodes:
- Modified `aWsm` to write the global `globals_len` containing the number of `(global)` instructions in the source wasm module
- Extended this through the SLEdge ABI interfaces
- Uses the value to size the globals vector during sandbox initialization.

### DIFF
--- a/runtime/include/sledge_abi_symbols.h
+++ b/runtime/include/sledge_abi_symbols.h
@@ -16,6 +16,7 @@ struct sledge_abi_symbols {
 	sledge_abi__entrypoint_fn_t   entrypoint;
 	uint32_t                      starting_pages;
 	uint32_t                      max_pages;
+	uint32_t                      globals_len;
 };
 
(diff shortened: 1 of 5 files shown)


runtime/include/sledge_abi_symbols.h
```diff
@@ -78,6 +79,14 @@ sledge_abi_symbols_init(struct sledge_abi_symbols *abi, char *path)
 	}
 	abi->max_pages = get_max_pages();
 
+	sledge_abi__wasm_memory_globals_len_fn_t get_globals_len = dlsym(abi->handle, SLEDGE_ABI__GLOBALS_LEN);
+	if (get_max_pages == NULL) {
```
emil916, line 83: `get_globals_len==NULL` instead?

emil916 (changes requested):

emil916 (approved):
LGTM!

PR #318, opened 2022-03-22
feat: sigfpe and sigsegv wasm traps

bushidocodes:
Replaces the conditionals to check for DIVZERO or overflow with a SIGFPE trap handler that triggers a wasm trap if running a sandbox in `RUNNING_USER` state.

The `libsledge` component still has conditionals for truncation instructions. For example:

```c
uint32_t
u32_trunc_f32(float f)
{
	if (unlikely(f < 0 || f > (float)UINT32_MAX))
		sledge_abi__wasm_trap_raise(WASM_TRAP_ILLEGAL_ARITHMETIC_OPERATION);
	return (uint32_t)f;
}
```

I don't believe there is a way to handle these checks with a trap handler, but I wanted to bring it up in case you have any ideas.


runtime/include/current_sandbox.h
```diff
@@ -107,5 +107,5 @@ current_sandbox_trap(enum sledge_abi__wasm_trap trapno)
 	assert(sandbox != NULL);
 	assert(sandbox->state == SANDBOX_RUNNING_USER || sandbox->state == SANDBOX_RUNNING_SYS);
 
-	longjmp(sandbox->ctxt.start_buf, trapno);
+	siglongjmp(sandbox->ctxt.start_buf, trapno);
```
gparmer, line 110: good catch!

libsledge/src/numeric_instructions.c
```diff
@@ -54,60 +54,48 @@ rotr_u64(uint64_t n, uint64_t c_u64)
...
 u64_rem(uint64_t a, uint64_t b)
 {
-	if (unlikely(b == 0)) sledge_abi__wasm_trap_raise(WASM_TRAP_ILLEGAL_ARITHMETIC_OPERATION);
 	return a % b;
 }
 
 INLINE int64_t
 i64_div(int64_t a, int64_t b)
 {
-	if (unlikely(b == 0 || (a == INT64_MIN && b == -1)))
```
gparmer, line 101: I don't understand the second condition here. If `a == INT64_MIN && b == -1` will this always cause a signal?

runtime/include/software_interrupt.h
```diff
@@ -31,7 +31,7 @@ software_interrupt_mask_signal(int signal)
 	sigset_t set;
 	int      return_code;
 
-	assert(signal == SIGALRM || signal == SIGUSR1);
+	assert(signal == SIGALRM || signal == SIGUSR1 || signal == SIGFPE || signal == SIGSEGV);
```
gparmer, line 34: Nice defensive coding

runtime/include/software_interrupt_counts.h
```diff
@@ -10,6 +10,8 @@ extern _Atomic volatile sig_atomic_t *software_interrupt_counts_deferred_sigalrm
 extern _Atomic volatile sig_atomic_t *software_interrupt_counts_deferred_sigalrm_replay;
 extern _Atomic volatile sig_atomic_t *software_interrupt_counts_sigalrm_kernel;
 extern _Atomic volatile sig_atomic_t *software_interrupt_counts_sigalrm_thread;
+extern _Atomic volatile sig_atomic_t *software_interrupt_counts_sigfpe;
```
gparmer, line 13: Love it

runtime/include/software_interrupt_counts.h
```diff
@@ -10,6 +10,8 @@ extern _Atomic volatile sig_atomic_t *software_interrupt_counts_deferred_sigalrm
...
+extern _Atomic volatile sig_atomic_t *software_interrupt_counts_sigsegv;
```
gparmer, line 14: Oh no! These are not cache aligned. This is a large amount of false sharing that will cause a big perf decrease on multicore.

We want to essentially have each atomic be 128 bytes apart (on 128-aligned boundaries). The typical way to accomplish this without blowing up the cache  and memory consumption is to put the counters in a struct, and align the structs.

```c
struct event_counts {
    _Atomic volatile sig_atomic_t software_interrupt_counts_sigalarm_reply;
    _Atomic volatile sig_atomic_t software_interrupt_counts_kernel;
    /* ... */     
} __attribute((aligned(128)));

struct event_counts software_interrupt_counts[runtime_worker_threads_count];
```

runtime/src/current_wasm_module_instance.c
```diff
@@ -27,5 +27,5 @@ sledge_abi__current_wasm_module_instance_trap(enum sledge_abi__wasm_trap trapno)
 	assert(sandbox != NULL);
 	assert(sandbox->state == SANDBOX_RUNNING_USER || sandbox->state == SANDBOX_RUNNING_SYS);
 
-	longjmp(sandbox->ctxt.start_buf, trapno);
+	siglongjmp(sandbox->ctxt.start_buf, trapno);
```
gparmer, line 30: I'm not sure I can build the context to understand all of this right now, and I think that @emil916 should understand the control flow/intention here. 

Looks good to me, but I suggest that he eventually validate.

runtime/src/software_interrupt.c
```diff
@@ -149,6 +149,30 @@ software_interrupt_handle_signals(int signal_type, siginfo_t *signal_info, void
...
 		break;
 	}
+	case SIGFPE: {
+		software_interrupt_counts_sigfpe_increment();
+
+		if (likely(current_sandbox && current_sandbox->state == SANDBOX_RUNNING_USER)) {
+			atomic_fetch_sub(&handler_depth, 1);
+			current_sandbox_trap(WASM_TRAP_ILLEGAL_ARITHMETIC_OPERATION);
+		} else {
+			panic("Runtime SIGFPE\n");
```
gparmer, line 159: Great job catching this case.

runtime/src/current_wasm_module_instance.c
gparmer: I'd imagine that you're reporting/recording the failure at the `setjmp` site.

gparmer (approved):
This looks brilliant. Great job. The error handling path for all of this is challenging, and you engineered it quite well.

libsledge/src/numeric_instructions.c
bushidocodes: I believe the second condition is an overflow because under two's complement, the negative side can represent one more number than the positive side.

bushidocodes: I changed the `traps` app so that the numerator was INT32_MIN to test this, and it seemed to trap. I temporarily made this a long to confirm that longs behave the same way. On Intel at least, both seem to trap with a SIGFPE. I guess that means it's not just floating-point units that trigger this.

runtime/include/software_interrupt_counts.h
bushidocodes: Thanks for letting me know! This is new to me. I'll push a commit with a fix.

runtime/src/current_wasm_module_instance.c
bushidocodes: Pretty straight-forward.

```c
void
current_sandbox_start(void)
{
	struct sandbox *sandbox = current_sandbox_init();

	int rc = setjmp(sandbox->ctxt.start_buf);
	if (rc == 0) {
		struct module *current_module = sandbox_get_module(sandbox);
		sandbox->return_value         = module_entrypoint(current_module);
	} else {
		current_sandbox_wasm_trap_handler(rc);
	}

	current_sandbox_fini();
}
```

```c
void
current_sandbox_wasm_trap_handler(int trapno)
{
	char           *error_message = NULL;
	struct sandbox *sandbox       = current_sandbox_get();
	sandbox_syscall(sandbox);

	switch (trapno) {
	case WASM_TRAP_EXIT:
		break;
	case WASM_TRAP_INVALID_INDEX:
		error_message = "WebAssembly Trap: Invalid Index\n";
		client_socket_send(sandbox->client_socket_descriptor, http_header_build(500), http_header_len(500),
		                   current_sandbox_sleep);
		break;
	case WASM_TRAP_MISMATCHED_TYPE:
		error_message = "WebAssembly Trap: Mismatched Type\n";
		client_socket_send(sandbox->client_socket_descriptor, http_header_build(500), http_header_len(500),
		                   current_sandbox_sleep);
		break;
	case WASM_TRAP_PROTECTED_CALL_STACK_OVERFLOW:
		error_message = "WebAssembly Trap: Protected Call Stack Overflow\n";
		client_socket_send(sandbox->client_socket_descriptor, http_header_build(500), http_header_len(500),
		                   current_sandbox_sleep);
		break;
	case WASM_TRAP_OUT_OF_BOUNDS_LINEAR_MEMORY:
		error_message = "WebAssembly Trap: Out of Bounds Linear Memory Access\n";
		client_socket_send(sandbox->client_socket_descriptor, http_header_build(500), http_header_len(500),
		                   current_sandbox_sleep);
		break;
	case WASM_TRAP_ILLEGAL_ARITHMETIC_OPERATION:
		error_message = "WebAssembly Trap: Illegal Arithmetic Operation\n";
		client_socket_send(sandbox->client_socket_descriptor, http_header_build(500), http_header_len(500),
		                   current_sandbox_sleep);
		break;
	}

	debuglog("%s", error_message);
	sandbox_close_http(sandbox);
	generic_thread_dump_lock_overhead();
	current_sandbox_exit();
	assert(0);
}
```

```c
void
current_sandbox_exit()
{
	struct sandbox *exiting_sandbox = current_sandbox_get();
	assert(exiting_sandbox != NULL);

	generic_thread_dump_lock_overhead();

	switch (exiting_sandbox->state) {
	case SANDBOX_RETURNED:
		sandbox_exit_success(exiting_sandbox);
		break;
	case SANDBOX_RUNNING_SYS:
		sandbox_exit_error(exiting_sandbox);
		break;
	default:
		panic("Cooperatively switching from a sandbox in a non-terminal %s state\n",
		      sandbox_state_stringify(exiting_sandbox->state));
	}

	scheduler_cooperative_sched(true);

	/* The scheduler should never switch back to completed sandboxes */
	assert(0);
}
```

libsledge/src/numeric_instructions.c
gparmer: Awesome!

gparmer: yeah, I think that sigfpe includes div by zero, which is applicable for integers as well.

runtime/include/software_interrupt_counts.h
gparmer: Yup, happy to chat about it at any point.

runtime/src/current_wasm_module_instance.c
gparmer: Sorry, stale comment. I ended up chasing this path down. Look fantastic.

PR #334, opened 2022-04-13
Json refactor

bushidocodes:
The reason I said this isn't quite ready for adding the new keys is that I anticipate that a module object will have some keys that contain array of objects. For example, this might be a "exports" array that contains export objects with a "symbol", a "path", and some scheduler logic. We can't do this currently because our parser assumes that we only have flat module objects. I suggest I handle that as a follow-on effort because that parsing logic change might benefit from a more focused review.

The main changes here are that I refactored the huge `module_alloc_from_json` function into several tiers:
- Loading a *.json file into a in-memory buffer 
- JSON parsing of an in-memory buffer
- Module allocation / initialization
- Module Listening

I added a bunch of JSON validation logic and error handling and did a bit of manual testing. Essentially, the runtime logs the error, dumps the JSON text, and then panics on these sorts of errors.

I also brought in the `module_database` code that has been dormant and unused for like two years! 

Some of the dynamic memory allocation is a bit complicated. I suggest that might be a good thing to focus on during the review.

### DIFF
--- a/runtime/include/json.h
+++ b/runtime/include/json.h
@@ -1,0 +1,230 @@
+#pragma once
+
+#include <assert.h>
+#include <stdlib.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <string.h>
+#include <jsmn.h>
+
+#include "runtime.h"
+#include "http.h"
+#include "module.h"
+
+static const int JSON_MAX_ELEMENT_COUNT = 16;
+static const int JSON_MAX_ELEMENT_SIZE  = 1024;
+
+struct module_config {
+	char    *name;
+	char    *path;
+	int      port;
+	uint32_t expected_execution_us;
+	int      admissions_percentile;
+	uint32_t relative_deadline_us;
+	int32_t  http_req_size;
+	int32_t  http_resp_size;
+	char    *http_resp_content_type;
+};
+
+static void
+print_module_config(struct module_config *config)
+{
+	printf("Name: %s\n", config->name);
+	printf("Path: %s\n", config->path);
+	printf("Port: %d\n", config->port);
+	printf("expected_execution_us: %u\n", config->expected_execution_us);
+	printf("admissions_percentile: %d\n", config->admissions_percentile);
+	printf("relative_deadline_us: %u\n", config->relative_deadline_us);
+	printf("http_req_size: %u\n", config->http_req_size);
+	printf("http_resp_size: %u\n", config->http_resp_size);
+	printf("http_resp_content_type: %s\n", config->http_resp_content_type);
+}
+
+static inline char *
+jsmn_type(jsmntype_t type)
+{
+	switch (type) {
+	case JSMN_UNDEFINED:
+		return "Undefined";
+	case JSMN_OBJECT:
+		return "Object";
+	case JSMN_ARRAY:
+		return "Array";
+	case JSMN_STRING:
+		return "String";
+	case JSMN_PRIMITIVE:
+		return "Primitive";
+	default:
+		return "Invalid";
+	}
+}
+
+/**
+ * Parses a JSON file into an array of module configs
+ * @param file_name The path of the JSON file
+ * @return module_config_vec_len on success. -1 on Error
+ */
+static inline int
+parse_json(const char *json_buf, ssize_t json_buf_size, struct module_config **module_config_vec)
+{
+	assert(json_buf != NULL);
+	assert(json_buf_size > 0);
+	assert(module_config_vec != NULL);
+
+	jsmntok_t tokens[JSON_MAX_ELEMENT_SIZE * JSON_MAX_ELEMENT_COUNT];
+	int       module_config_vec_len = 0;
+
+
+	/* Initialize the Jasmine Parser and an array to hold the tokens */
+	jsmn_parser module_parser;
+	jsmn_init(&module_parser);
+
+	/* Use Jasmine to parse the JSON */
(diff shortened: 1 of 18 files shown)


runtime/src/main.c
```diff
@@ -344,6 +347,77 @@ check_versions()
...
+	/* We will retry fclose when we fall through into stat_buffer_alloc_err */
+fread_err:
+	free(*file_buffer);
+stat_buffer_alloc_err:
+	// Check to ensure we haven't already close this
+	if (module_file != NULL) {
+		if (fclose(module_file) == EOF) panic("Failed to close file\n");
+	}
+err:
+	return (ssize_t)-1;
```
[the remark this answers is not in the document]
emil916: I see, since it can't be a negative number, I presume... But why not just zero? Since the size can't be zero anyways? Because (ssize_t)-1 can be an actual legitimate size, too. No?

runtime/include/json.h
```diff
@@ -0,0 +1,284 @@
...
+	int       module_config_vec_len = 0;
+
+
+	/* Initialize the Jasmine Parser and an array to hold the tokens */
+	jsmn_parser module_parser;
+	jsmn_init(&module_parser);
+
+	/* Use Jasmine to parse the JSON */
+	int total_tokens = jsmn_parse(&module_parser, json_buf, json_buf_size, tokens,
+	                              sizeof(tokens) / sizeof(tokens[0]));
```
emil916, line 158: Why not use `JSON_MAX_ELEMENT_SIZE * JSON_MAX_ELEMENT_COUNT` again instead of `sizeof(tokens) / sizeof(tokens[0])` ?

emil916: Actually why not just declare a single constant (e.g. JSON_MAX_SIZE=16*1024) that is equal to the product of JSON_MAX_ELEMENT_SIZE * JSON_MAX_ELEMENT_COUNT ? Since you are using them only once and multiplying anyways.

runtime/src/main.c
```diff
@@ -380,8 +454,37 @@ main(int argc, char **argv)
...
+			exit(-1);
+		}
+
+		/* Start listening for requests */
+		rc = module_listen(module);
+		if (rc < 0) exit(-1);
+	}
+
+	for (int module_idx = 0; module_idx < module_config_vec_len; module_idx++) {
+		module_config_deinit(&module_config_vec[module_idx]);
```
emil916, line 485: `free(&module_config_vec[module_idx])` after deinit?

runtime/include/json.h
```diff
@@ -0,0 +1,284 @@
...
+	if (module_config_vec_len == 0) {
+		fprintf(stderr, "Config is an empty JSON array\n");
+		goto err;
+	}
+
+	*module_config_vec = (struct module_config *)calloc(module_config_vec_len, sizeof(struct module_config));
+	int module_idx     = -1;
+	int module_fields_remaining = 0;
+
+	for (int i = 1; i < total_tokens; i++) {
```
emil916, line 190: Why start from `i=1`? This is `jsmn` thing?

runtime/include/json.h
```diff
@@ -0,0 +1,284 @@
...
+
+			sprintf(key, "%.*s", tokens[i].end - tokens[i].start, json_buf + tokens[i].start);
+
+			/* Validate that key has a value */
+			if (!has_valid_size(tokens[i], key, 1)) goto json_parse_err;
+
+			/* Advance to Value */
+			i++;
+
+			if (strcmp(key, "name") == 0) {
```
emil916, line 213: I feel like the key names should be wrapped up in a global const array, near the `module_config` struct, for example.

emil916 (commented):
Cool work! Left some comments. Please have a look.

runtime/include/json.h
```diff
@@ -0,0 +1,284 @@
...
+		return "Primitive";
+	default:
+		return "Invalid";
+	}
+}
+
+static inline bool
+has_valid_size(jsmntok_t tok, char *key, int expected_size)
+{
+	if (tok.size != 1) {
```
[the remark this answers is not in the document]
bushidocodes: Good catch! You're right!

runtime/include/json.h
```diff
@@ -0,0 +1,284 @@
...
+		} else {
+			/* Inside Object */
+
+			/* Validate that key is non-emptry string */
+			if (!is_valid_key(tokens[i])) goto json_parse_err;
+
+			sprintf(key, "%.*s", tokens[i].end - tokens[i].start, json_buf + tokens[i].start);
+
+			/* Validate that key has a value */
+			if (!has_valid_size(tokens[i], key, 1)) goto json_parse_err;
```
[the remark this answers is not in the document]
bushidocodes: `jsmn_tok_t` is a jasmine token. The jasmine data structure is effectively a tree and the `size` member is the number of children. The size of a JSON object is the number of keys. A key should only ever have one child: it's value. The size of a JSON array is the number of elements.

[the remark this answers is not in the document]
bushidocodes: Hitting resolved based on discussion in slack

runtime/include/json.h
bushidocodes: It's just because we've already checked that the 0th element was a JSON array above this.

runtime/src/main.c
bushidocodes: It's an array of structures, not pointers. See `json.h:184`

```c
*module_config_vec = (struct module_config *)calloc(module_config_vec_len, sizeof(struct module_config));
```

emil916 (approved):
Great job!

PR #336, opened 2022-04-28
Multi Tenancy with Deferrable Server

emil916:
This is a Multi Tenancy policy with Deferrable Server implementation. I tried to add the new policy as "pluggable" as possible. Still, there needs to be some soft engineering on making the process of adding a new policy completely pluggable.
It is huge, but the major important changes are in the source codes. So, you can skim through the shell scripts and other experiment files.


runtime/include/json.h
```diff
@@ -254,6 +258,19 @@ parse_json(const char *json_buf, ssize_t json_buf_size, struct module_config **m
 				int rc = parse_uint32_t(tokens[i], json_buf, module_keys[module_expected_execution_us],
 				                        &(*module_config_vec)[module_idx].expected_execution_us);
 				if (rc < 0) goto json_parse_err;
+			} else if (strcmp(key, module_keys[module_replenishment_period_us]) == 0) {
```
gparmer, line 261: Unfortunately, I'm sure this will all conflict with @bushidocodes updates to the json.

runtime/include/module.h
```diff
@@ -62,6 +88,14 @@ struct module {
 	struct admissions_info admissions_info;
 	uint64_t               relative_deadline; /* cycles */
 
+	/* Deferrable Server Attributes */
+	uint64_t                 replenishment_period; /* cycles, not changing after init */
+	uint64_t                 max_budget;           /* cycles, not changing after init */
+	_Atomic volatile int64_t remaining_budget;     /* cycles left till next replenishment, can be negative */
+
+	struct perworker_module_sandbox_queue *pwm_sandboxes;
```
gparmer, line 96: No idea what pwm stands for. Normally it is pulse-width modulated...

runtime/include/priority_queue.h
```diff
@@ -332,7 +331,7 @@ static inline int
 priority_queue_length_nolock(struct priority_queue *priority_queue)
 {
 	assert(priority_queue != NULL);
-	assert(!listener_thread_is_running());
+	if (scheduler != SCHEDULER_MTDS) assert(!listener_thread_is_running());
```
gparmer, line 334: By moving this out of the assert, you're losing the branch prediction hint.  Maybe should be:

```c
assert(scheduler == SCHEDULER_MTDS || !listener_thread_is_running());
```

runtime/include/priority_queue.h
```diff
@@ -404,7 +403,7 @@ priority_queue_delete_nolock(struct priority_queue *priority_queue, void *value)
 {
 	assert(priority_queue != NULL);
 	assert(value != NULL);
-	assert(!listener_thread_is_running());
+	if (scheduler != SCHEDULER_MTDS) assert(!listener_thread_is_running());
```
gparmer, line 406: same. Similar below.

runtime/include/scheduler.h
```diff
@@ -122,6 +162,8 @@ static inline struct sandbox *
 scheduler_get_next()
 {
 	switch (scheduler) {
+	case SCHEDULER_MTDS:
```
gparmer, line 165: If this list continues to grow, doing dispatch with switches won't scale well at all. Maybe a concern for the future, but it would be good to understand how long it takes to schedule to make a more fact-based decision.

runtime/src/current_sandbox.c
```diff
@@ -212,6 +212,9 @@ current_sandbox_fini()
 	assert(sandbox->state == SANDBOX_RUNNING_SYS);
 	sandbox_close_http(sandbox);
 	sandbox_set_as_returned(sandbox, SANDBOX_RUNNING_SYS);
+	if (module_is_paid(sandbox->module)) {
```
gparmer, line 215: You have replicated this code around 5 times. Seems like in the long term, it should be a function.

gparmer (commented):
Sorry, but I don't have much context in the system for this. My comments are very "surface level" and not very useful.

runtime/include/json.h
emil916: Yes.

runtime/include/module.h
emil916: It is an acronym following the struct name "per worker module". It is ugly, though, I know.

runtime/include/module.h
```diff
@@ -186,6 +220,41 @@ module_release(struct module *module)
...
+ * Compute the next timeout given a module's replenishment period
+ * @param m_replenishment_period
+ * @return given module's next timeout
+ */
+static inline uint64_t
+get_next_timeout_of_module(uint64_t m_replenishment_period)
+{
+	uint64_t now = __getcycles();
+	return runtime_boot_timestamp
+	       + ((now - runtime_boot_timestamp) / m_replenishment_period + 1) * m_replenishment_period;
```
bushidocodes, line 255: I suspect that this might be better if the initial timestamp represents when the tenant/module is actually up and able to receive requests. I believe that `runtime_boot_timestamp` occurs quite a bit earlier than that before the JSON is even parsed. Not really a big deal I suppose. Just makes the initial period a sort of random duration.

PR #340, opened 2022-05-05
chore: Assorted bash experiment cleanup

bushidocodes:
Ported over what was most clear to me. There are some other bash framework changes that seem like they'd cause some problems in some test configurations (running multiple scheduler options or running client and server on different machines).

We might as well merge this first if this makes sense and then break the rest of your bash changes into a different PR to discuss / test further.


bushidocodes:
@emil916 - Going to just merge this stuff since it's a subset of your code and it looked good to me.

emil916 (commented):
all good.

PR #344, opened 2022-05-23
feat: Initial scratch storage implementation

bushidocodes:


applications/Makefile
```diff
@@ -22,7 +22,11 @@ all: \
 	gocr.install \
 	gps_ekf.install \
 	license_plate_detection.install \
-	resize_image.install
+	resize_image.install \
+	scratch_storage_get.install \
```
gparmer, line 26: A little strange to have these apps in `wasm_apps` as they are quite sledge specific, but I'm sure that applies to multiple apps there. No problem here, just a little cognitive dissonance.

runtime/include/map.h
```diff
@@ -0,0 +1,183 @@
...
+#include <stdint.h>
+#include <stdlib.h>
+
+#include "lock.h"
+
+/* Simple K-V store based on The Practice of Programming by Kernighan and Pike */
+
+/* Bucket count is sized to be a prime that is approximately 20% larger than the desired capacity (6k keys) */
+#define MAP_BUCKET_COUNT 7907
+#define MAP_LOCK_STRIPES 17
```
gparmer, line 14: Great that you already instrumented the lock code (from before) to report total blocking time so that we can know if we have too much contention, and performance is paying. I think it is more than enough information to realize that there is a perf problem that requires debugging.

Nothing actionable: just a pat on the back of your past self.

runtime/include/map.h
```diff
@@ -0,0 +1,183 @@
...
+	uint32_t         hash;
+	uint8_t         *key;
+	uint32_t         key_len;
+	uint8_t         *value;
+	uint32_t         value_len;
+	struct map_node *next;
+};
+
+struct map {
+	struct map_node *buckets[MAP_BUCKET_COUNT];
```
gparmer, line 28: A downside of this structure is that we're going to have a cache-line bounce for the lock, and a *separate* cache line bounce for the budget. As MCS locks are small (only a word?), inlining it into the `map_node` wouldn't take too much memory. Not sure where the trade-offs pay off here. I dont' think that we're as concerned about memory as java's maps as we have many fewer of these than they do.

runtime/include/map.h
```diff
@@ -0,0 +1,183 @@
...
+#define MAP_HASH jenkins_hash
+
+struct map_node {
+	uint32_t         hash;
+	uint8_t         *key;
+	uint32_t         key_len;
+	uint8_t         *value;
+	uint32_t         value_len;
+	struct map_node *next;
+};
```
gparmer, line 25: The ordering of these will waste 96 bytes given alignment rules. Not sure if it is that much less readable if you pt the 32 bit values contiguously.

runtime/include/map.h
```diff
@@ -0,0 +1,183 @@
...
+	uint32_t hash = MAP_HASH(key, key_len);
+	LOCK_LOCK(&map->locks[hash % MAP_LOCK_STRIPES]);
+	for (struct map_node *node = map->buckets[hash % MAP_BUCKET_COUNT]; node != NULL; node = node->next) {
+		if (node->hash == hash) {
+			node->value_len = value_len;
+			node->value     = realloc(node->value, value_len);
+			assert(node->value);
+			memcpy(node->value, value, value_len);
+		}
+		goto DONE;
```
gparmer, line 160: I'm pretty confused how this code is expected to iterate multiple times. Have you tested with collisions?

runtime/include/map.h
```diff
@@ -0,0 +1,183 @@
...
+	}
+
+	struct map_node *new_node = (struct map_node *)malloc(sizeof(struct map_node));
```
gparmer, line 163: Too late to care about this now, but in systems that don't want to check the return value for `malloc`, they often define a `xmalloc` (I don't know why `x`) that will panic out with an error message if `malloc` returns `NULL`. Better to fail early.

runtime/src/sledge_abi.c
```diff
@@ -1039,3 +1040,126 @@ sledge_abi__wasi_snapshot_preview1_sock_shutdown(__wasi_fd_t fd, uint32_t how)
...
+ * @returns value_len at key or 0 if key not present
+ */
+EXPORT uint32_t
+sledge_abi__scratch_storage_get_size(uint32_t key_offset, uint32_t key_len)
+{
+	struct sandbox *sandbox = current_sandbox_get();
+
+	sandbox_syscall(sandbox);
+
+	uint8_t *key = (uint8_t *)get_memory_ptr_for_runtime(key_offset, key_len);
```
gparmer, line 1056: Can't this error out if they pass an offset beyond linear memory bounds? I'd imagine I'm wrong, and instead you just panic out of this specific sandbox in that case (`setjmp`).

gparmer (approved):
Nothing here requires a response. There looks to be a single bug with a trivial fix. Take all of this as input to your own trade-offs, and commit when ready!

runtime/include/map.h
gparmer: :point_up: this is the bug.

bushidocodes: woof, that's embarrassing!

bushidocodes: woof, that's embarrassing!

runtime/include/map.h
bushidocodes: I added an `xmalloc` and used it in this file. Will at least be available for future use.

bushidocodes: I added an `xmalloc` and used it in this file. Will at least be available for future use.

runtime/include/map.h
bushidocodes: I added a new struct `map_bucket` that contains a lock and a pointer to the head node. My main concern was the overall complexity of adding a lock on each node. That seems to suggest hand-over-hand locking to me, which seems like it'd be overly expensive assuming collisions are rare.

bushidocodes: I added a new struct `map_bucket` that contains a lock and a pointer to the head node. My main concern was the overall complexity of adding a lock on each node. That seems to suggest hand-over-hand locking to me, which seems like it'd be overly expensive assuming collisions are rare.

PR #351, opened 2022-07-12
General Cleanup

emil916:
Not a big one. Pretty straightforward:
- refactor: use var across the Makefile instead of using **sledgert** string

refactor bash libraries
- remove hash symbol for **csv_to_dat**
- add **set_print_pretty** for gdb mode
- add logging (log.txt) for client
- change printf format for perf_log in table from float to int since we will be using usec

cleanup: generalize redundant files in the experiments:
- remove individual gitignore, have single gitignore in parent (tests)
- remove individual env files, have single copies in 'common' dir
- remove individual install.sh, have install_tools.sh (in bash libs)


bushidocodes:
@emil916 - Looks like some tests are failing. I'll take a look once the checkmarks are green.

emil916:
@bushidocodes all green. Have a glance when you get a chance. Estimated time: 10m.

runtime/Makefile
```diff
@@ -128,13 +128,13 @@ all: thirdparty runtime
 clean: thirdparty.clean runtime.clean
 
 # sledgert Rules
-bin/sledgert: ${HEADER_DEPS} ${CFILES}
+bin/${BINARY_NAME}: ${HEADER_DEPS} ${CFILES}
```
bushidocodes, line 131: Nice catch!

bushidocodes (approved):
Looks like a good cleanup effort. Appreciate this!

PR #357, opened 2022-08-11
fix: lock tracking

### LINKED ISSUE #82: Track Time Waiting for Locks
Places where runtime faces lock contention:
* spins in ck_spinlock_fas_lock, called by priority_queue_length, priority_queue_enqueue, priority_queue_delete
* These are called by local_runqueue_minheap_add, global_request_scheduler_minheap_add, local_runqueue_minheap_delete

What about trylock?

bushidocodes:
Resolves #82


emil916:
Looks great. But can we associate these measurements per actual locks introduced in the system (instead of per thread). So, what I mean by this is for every lock in the system (see below) track the four metrics you introduced.  

- global runqueue lock
- local runqueue locks
- obj pool lock
- perf_window locks
- map
any other I might be forgetting?

emil916 (changes requested):
Looks great. But can we associate these measurements per actual locks introduced in the system (instead of per thread). So, what I mean by this is for every lock in the system (see below) track the four metrics you introduced.

- global runqueue lock
- local runqueue locks
- obj pool lock
- perf_window locks
- map

any other I might be forgetting?

emil916 (approved):
Looks like what I wanted! Thank you, @bushidocodes !

PR #358, opened 2022-08-18
fix: correct HTTP session state machine

bushidocodes:
This is a fix for a bug I discovered when running the CMU SOD app, which has a sufficiently large response body to hit our EWOULDBLOCK path. There was a bad assumption in the state machine. I cleaned up the states a bit and reworked the logic for the state transitions when sending a response.


emil916:
> There was a bad assumption in the state machine.
Can you please elaborate what that bad assumption was?

emil916 (approved):
Nevermind the previous comment. Got it. LGTM.

PR #363, opened 2022-08-29
refactor: HTTP memstreams

bushidocodes:
Refactors HTTP request and response buffers from the vec macro to the POSIX open_memstream interface.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,6 @@ An SLEdge serverless function consists of a shared library (\*.so) and a JSON co
   "port": 10000,
   "expected-execution-us": 600,
   "relative-deadline-us": 2000,
-  "http-req-size": 1024,
   "http-resp-content-type": "text/plain"
 }
 ```
--- a/README.md
+++ b/README.md
@@ -87,7 +87,6 @@ An SLEdge serverless function consists of a shared library (\*.so) and a JSON co
   "expected-execution-us": 600,
   "relative-deadline-us": 2000,
   "http-req-size": 1024,
-  "http-resp-size": 1024,
   "http-resp-content-type": "text/plain"
 }
 ```
--- a/runtime/include/http_session.h
+++ b/runtime/include/http_session.h
@@ -186,8 +186,9 @@ http_session_set_response_header(struct http_session *session, int status_code)
 	assert(rc != EOF);
 
 	if (status_code == 200) {
+		/* Make sure the response_body is flushed */
 		int rc = auto_buf_flush(&session->response_body);
-		if (unlikely(rc != 0)) { panic("response_header auto_buf failed to flush: %s\n", strerror(errno)); };
+		if (unlikely(rc != 0)) { panic("response_body auto_buf failed to flush: %s\n", strerror(errno)); };
 
 		/* Technically fprintf can truncate, but I assume this won't happen with a memstream */
 		rc = fprintf(session->response_header.handle, HTTP_RESPONSE_CONTENT_TYPE,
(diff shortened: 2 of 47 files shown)


README.md
```diff
@@ -87,7 +87,6 @@ An SLEdge serverless function consists of a shared library (\*.so) and a JSON co
   "expected-execution-us": 600,
   "relative-deadline-us": 2000,
   "http-req-size": 1024,
```
emil916, line 89: This should also be gone now, I guess.

runtime/include/http_session.h
```diff
@@ -187,20 +182,30 @@ http_session_set_response_header(struct http_session *session, int status_code,
...
-		                                           HTTP_RESPONSE_200_TEMPLATE, content_type, content_length);
-	} else {
-		size_t header_len = http_header_len(status_code);
-		size_t to_copy    = HTTP_SESSION_RESPONSE_HEADER_CAPACITY < header_len
-		                      ? HTTP_SESSION_RESPONSE_HEADER_CAPACITY
-		                      : header_len;
-
-		strncpy(session->response_header, http_header_build(status_code), to_copy);
-		session->response_header_length = to_copy;
+		int rc = auto_buf_flush(&session->response_body);
```
emil916, line 189: shouldn't this be `session->response_header` instead?

runtime/include/http_session.h
```diff
@@ -419,11 +371,21 @@ http_session_receive_request(struct http_session *session, void_star_cb on_eagai
...
+		fflush(session->request_buffer.handle);
 
-		ssize_t bytes_parsed = http_session_parse(session, bytes_received);
+		/* Invalidate parser structure if buffer moved */
+		ssize_t bytes_parsed;
+		if (old_buffer != session->request_buffer.data) {
+			http_session_parser_init(session);
+			bytes_parsed = http_session_parse(session, session->request_buffer.size);
+		} else {
+			bytes_parsed = http_session_parse(session, bytes_received);
```
emil916, line 387: Looking at the implementation of http_session_parse the second argument seems to be useless. So, this if clause go away ?

runtime/src/sandbox.c
```diff
@@ -94,7 +94,7 @@ sandbox_prepare_execution_environment(struct sandbox *sandbox)
 
 	int rc;
 
-	rc = http_session_init_response_buffer(sandbox->http, sandbox->route->response_size);
+	rc = http_session_init_response_body(sandbox->http);
 	if (rc < 0) {
 		error_message = "failed to allocate response buffer";
```
emil916, line 99: failed to allocate response "body"

runtime/include/http_session.h
emil916, line 387: Wondering if we could just flush & http_parse only once at the end of receiving all...

runtime/include/http_session.h
```diff
@@ -456,26 +418,10 @@ static inline int
 http_session_write_response(struct http_session *session, const uint8_t *source, size_t n)
```
emil916, line 418: This func became a zombie function now, since the only occurrence in "wasi_impl_serverless.c:813" was removed.

emil916: Reverted back the wasi_impl code, since the function has some assertions.

runtime/include/http_session.h
emil916: Actually this should be session->response_body, since we must be flushing it at some point before sending.

runtime/include/http_session.h
emil916: Ended up actually using it for the updated version of the http_session_parse function (line:304)

emil916: Ended up just updating http_parser structure when the buffer address moved by memstream.

PR #364, opened 2022-08-30
feat: Latency perflog

bushidocodes:
@emil916 requested that I add a perf_window to each route to capture the sandbox total times. This includes all blocking, waiting, etc. Be aware that the lock for the perf_window structure likely impacts performance to some degree. I think this is inevitable and probably okay for the demo. However, there might be room to make the perf_window structure more efficient in the future.


runtime/src/metrics_server.c
```diff
@@ -81,6 +66,21 @@ metrics_server_handler(void *arg)
 	/* Intermediate cast to integral value of 64-bit width to silence compiler nits */
 	int client_socket = (int)(long)arg;
 
+	/* Duplicate fd so fclose doesn't close the actual client_socket */
```
bushidocodes, line 69: I discovered a bug when testing. I was trying to route before spawning a thread, but fgets can block. This could take down the listener core. Moving this into the thread resolved the issue.

emil916: makes sense

emil916 (approved):
Approving with some changes to:
- the perf_window_get_percentile was not considering cases where perf_count<CAP
- changed latency unit from ms to us

PR #365, opened 2022-09-21
remove epoll from workers

emil916:
After the recent `http_session` changes made to the listener core to handle all the receive and send blockings, sandboxes running on worker cores now never "sleep", which means the worker cores never handle blocking on I/O anymore.

That is why keeping `epoll `support for workers seems meaningless now. The only epoll in Sledge is then the Listener core epoll.

P.S. There is one potential case we might need to bring back the epoll support back to workers. That is:

> If the client kills the connection before we're done computing, we won't know on the local worker. So, we either need epoll on each worker OR just use the listener core EPOLL and messaging mechanism between the Listener and Worker to inform the worker to cur off the disconnected work. @gparmer


emil916:
This is not a big change in code. It passes all the tests. @gparmer already approved this change. So, going to merge.

PR #366, opened 2022-09-24
Retrieve more accurate CPU frequency for runtime

emil916:
Till now we were for retrieving the CPU freq, we were getting the top value of the output of `/proc/cpuinfo`  , but because that is just the reserved core, it was giving inaccurate frequency, causing inaccurate sandbox logs later on.

With the new hack, I am getting the **mean average** of CPU freqs of all the active worker cores (NOT the reserved and the event cores). The reason for getting the average is that, upon investigating, I noticed how the cores on different sockets can have varying (not so much but still) CPU frequencies, particularly when not all the cores are set to be the workers.

Additionally, I applied a **toggle to turn on/off** the idle **spinloop pause** in the worker. It is OFF by default.


gparmer:
This looks fine, but I don't think it is solving the problem (if I understand it correctly).

I figure the main purpose of this is to be able to convert between cycles and real-time. If that's the case, most processors we use likely use "invariant tsc"s (or, at least "constant tsc"s) (e.g, see https://stackoverflow.com/questions/62492053/can-constant-non-invariant-tsc-change-frequency-across-cpu-states). 

In that case, the TSC cycles tick by independent of the CPU state and frequency. So if that's the goal, you really want to figure out the tick rate of the TSC, independent of CPU frequency.

gparmer (approved):
See comment in the main comment thread. Not sure this is solving the problem, thought it is better than what we had previously.

gparmer:
More info: https://forums.guru3d.com/threads/a-bit-detailed-info-on-intel-time-stamp-counter-tsc.433977/

gparmer:
An the most useful documentation I've found: https://blog.trailofbits.com/2019/10/03/tsc-frequency-for-all-better-profiling-and-benchmarking/

See the `/sys/devices/system/cpu/cpu0/tsc_freq_khz` file

emil916:
This is a great feedback! Thank you, Gabe!
I'll go ahead and merge this and look into tsc later for sure!

PR #367, opened 2022-09-26
remove sending duplicate http response header

lyuxiaosu:
Remove sending the duplicate http response header. This will cause **hey** failed. With **curl**, you will see the duplicate http response header:
```
HTTP/1.1 200 OK
Server: SLEdge
Connection: close
Content-Type: text/plain
Content-Length: 4
HTTP/1.1 200 OK
Server: SLEdge
Connection: close

144
```
Line 185 has already sent the http response header


emil916:
This is a great finding! Thank you for the fix, @lyuxiaosu !

PR #368, opened 2022-09-29
Remove wasm_apps submodule from Sledge repo

emil916:
The same wasm_apps repo was linked as a submodule to both aWsm and Sledge, which causes some redundancy issues in the workspace. 

In this PR I removed the wasm_apps submodule from Sledge and instead use awsm's copy as a symlink.


emil916:
Since it is a trivial cleanup PR, I am going to go ahead merge it to move on.

PR #369, opened 2022-10-03
fib errors for test scripts of fibonacci: Sending a http request with…

lyuxiaosu:
It seems Sledge right now only supports sending a http request with a http body by POST not GET. You can reproduce this by running hey command with `hey -disable-compression -disable-keepalive -disable-redirects -z "${duration_sec}s" -cpus 2 -c 100 -t 0 -o csv -m GET -d "29\n" "http://${hostname}:10010/fib"`, you will find Sledge reply 200 OK but 0 fib result. After replacing GET with POST, sledge can return the right result.


emil916:
Fixed in the latest master branch. Thank you!

PR #388, opened 2026-06-13
fix: repair Docker dev-environment build and clarify native setup

bushidocodes:
## Summary

The documented setup instructions were failing. The **native** path assumes a Debian host with **LLVM 13**, but aWsm's `llvm-alt` bindings call LLVM C-API functions (`LLVMBuildCall`, `LLVMBuildLoad`) that were **removed in LLVM 15+** — so on a newer host (e.g. Ubuntu 24.04 / LLVM 21) the build fails to link (`undefined symbol: LLVMBuildCall`), and `install_deb.sh` can't even run (`libtinfo5` and LLVM 13 aren't available there).

The **Docker** route exists precisely to pin the LLVM-13 toolchain, but it had bit-rotted in several places. This PR repairs the Docker build and documents the native constraint. (A separate change to the `gocr` zlib download URL is needed too, but it lives in a nested submodule and will be handled via the submodule chain.)

## Changes

**`Dockerfile.x86_64`**
- Install `hey` from focal's `universe` repo instead of downloading the prebuilt binary — the `hey-release` S3/GCS URLs now return **403**.
- `git config --system --add safe.directory '*'` so the root-run `make install` doesn't abort with **"dubious ownership"** on the host-owned, bind-mounted `/sledge`.
- Set `RUSTUP_HOME`/`CARGO_HOME` to the `dev` user's install so the root-run build can resolve the default Rust toolchain instead of failing with **"no default is configured"**.

**`Makefile`**
- Make the `wasm_apps` symlink rule idempotent (`ln -srfn`). The symlink is committed to the repo, so it already exists on a fresh clone and the old `ln -sr` failed with **"File exists"**, breaking `make install`.

**`README.md`**
- Recommend the Docker route and add an explicit caveat to the native section documenting the LLVM-13/Debian requirement and why Ubuntu 24.04 (and other non-LLVM-13 hosts) can't use it.

## Verification

`./devenv.sh setup` now completes end-to-end: aWsm, the runtime, and all benchmark apps build. The runtime was launched against the built binaries and served:
- **fibonacci** — `GET /fib?10` → `55`, `?20` → `6765` (correct HTTP headers)
- **gocr** — OCR output exact-matched the test's `expected_result.txt`
- **resize_image** — output JPEG byte-identical to `expected_result.jpg`

### DIFF
--- a/Dockerfile.x86_64
+++ b/Dockerfile.x86_64
@@ -1,5 +1,5 @@
-# using ubuntu 20 docker image
-FROM ubuntu:focal
+# using ubuntu 24.04 (noble) docker image
+FROM ubuntu:noble
 
 ENV LLVM_VERSION=13
 ENV WASI_SDK_VERSION=12
@@ -27,7 +27,6 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 	binutils-dev \
 	cmake \
 	git \
-	libtinfo5 \
 	libtool \
 	pkg-config
 
@@ -47,7 +46,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 	wget
 
 # Test Script Stuff
-# hey is a load generator used by the test workloads (focal ships it in universe;
+# hey is a load generator used by the test workloads (noble ships it in universe;
 # the previously-used hey-release S3/GCS binaries now return 403)
 RUN apt-get update && apt-get install -y --no-install-recommends \
 	bc \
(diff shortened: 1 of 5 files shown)


bushidocodes:
I'm potentially interested in using some of my AI tokens to get this up to date. Here's a PR of some issues I hit trying to clone and run the project.

I've also opened a PR against gocr due to a link not returning a 404. That will take a few layers of PRs to get merged here due to our multi-layer submodule setup.

Dockerfile.x86_64
emil916: This can be upgraded to "noble", I have been using it locally and it is compatible.

emil916 (changes requested):
Other than that it looks good.

Dockerfile.x86_64
bushidocodes: @emil916 - I pushed a new commit for this and tested the Docker and native install paths on Noble.

PR #389, opened 2026-06-16
fix: graceful TCP close to stop RST on early error responses (#185)

### LINKED ISSUE #185: Hey complains when it receives 503s.
When running `hey`, 503s result in the following being logged:

```sh
Unsolicited response received on idle HTTP channel starting with "HTTP/1.1 503 Service Unavailable\r\n\r\n"; err=<nil>
```

It is unclear if this is expected behavior or an indication that we are not properly sending 503s.

Run `./runtime/experiments/premption/run.sh` to recreate

bushidocodes:
Closes #185

## Summary

Fixes the root cause behind #185 — `hey` logging `Unsolicited response received on idle HTTP channel starting with "HTTP/1.1 503 ..."`.

SLEdge often answers a request before it has finished reading it: the short-circuit error responses (400/404/429/500/503) in the listener thread write the response and `close()` the socket while the client may still be sending its request body. `tcp_session_close()` did a bare `close()`, and on Linux a `close()` with unread data in the kernel receive buffer **discards it and emits a TCP RST** instead of a graceful FIN. The client surfaces that RST as `connection reset by peer` or, on a pooled keepalive connection, as the exact `Unsolicited response received on idle HTTP channel` warning in the issue. This matches the maintainers' own hypothesis in the issue thread ("we write a response before reading the entire HTTP request").

## Change

`tcp_session_close()` now closes gracefully:
1. `shutdown(SHUT_WR)` — send our FIN so the client (now holding a complete response) stops sending.
2. Drain already-buffered inbound data with a **non-blocking, bounded** loop (32 × 8 KiB) so `close()` never sees unread data, while guaranteeing the single listener thread can't block or spin on a slow/malicious client.

One file, +41 lines, no change to the success path.

## Verification

Reproduced and verified in the Docker dev environment with `hey`:

| Case | Before | After |
|---|---|---|
| Normal 200 traffic (5000 req) | 200×5000, ~15k rps | unchanged, 0 warnings |
| GET → 429/503 (originally reported scenario) | already clean | clean |
| **POST 100 KB body → 404 route** (`connection reset by peer`) | **~159 / 2400 req** | **0** |

Note: the originally reported bodyless-GET scenario is already clean on current `master` (admissions/scheduler 429s now fire *after* the full request is read). The reproducible residual was a request **with a body** hitting an early-error route, which this fixes.

## Known limitation

A client that keeps streaming a *large* body to a rejected route may still observe its own write abort as `broken pipe` (EPIPE). Fully suppressing that would require lingering on the request in the event loop (a draining state in the listener), which is disproportionate for this log-level issue — left out of scope and documented in the code comment.

### DIFF
--- a/repro_issue_185.sh
+++ b/repro_issue_185.sh
@@ -1,0 +1,179 @@
+#!/usr/bin/env bash
+#
+# repro_issue_185.sh — Reproduce the TCP-reset bug behind issue #185 / PR #389.
+#
+# Symptom: `hey`/clients see "connection reset by peer" (and pooled keepalive
+# clients log `Unsolicited response received on idle HTTP channel starting with
+# "HTTP/1.1 ..."`) when SLEdge answers a request *before* it has finished reading
+# the request body.
+#
+# Why a plain bodyless GET does NOT reproduce it:
+#   429/500/503 are emitted from on_client_request_received(), which only runs
+#   after the FULL request (body included) has been read. So those are clean.
+#
+# The reproducible path is 404. In on_client_request_receiving()
+# (runtime/src/listener_thread.c, ~line 200) the route is matched the moment the
+# URL is parsed from the *request line* — before the body arrives:
+#
+#     if (session->route == NULL && strlen(session->http_request.full_url) > 0) {
+#         route = http_router_match_route(...);
+#         if (route == NULL) { ...404...; on_client_response_header_sending(); return; }
+#     }
+#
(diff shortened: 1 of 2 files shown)


emil916 (commented):
I've never run into this issue. The server doesn't send any of the mentioned error responses (400/404/429/500/503), unless the request is fully received, no? I don't see how this is ever useful.

emil916:
I see. I remember now, I used to get these 503 errors back in 2022 when using hey (now I am pretty much using my own loadtest).

PR #390, opened 2026-06-16
Add Valgrind suppressions for runtime false positives

### LINKED ISSUE #69: Valgrind reports Invalid Read of Size 4 at software_interrupt_handle_signals
```
Number of cores 16, sandboxing cores 15 (start: 1) and module reqs 0
==8856== Thread 6:
==8856== Invalid read of size 4
==8856==    at 0x56132A6: pthread_kill (pthread_kill.c:40)
==8856==    by 0x409F66: software_interrupt_handle_signals (software_interrupt.c:72)
==8856==    by 0x561689F: ??? (in /lib/x86_64-linux-gnu/libpthread-2.27.so)
==8856==    by 0x56131F5: pthread_sigmask (pthread_sigmask.c:45)
==8856==    by 0x40A712: software_interrupt_unmask_signal (software_interrupt.h:86)
==8856==    by 0x40A712: worker_thread_main (worker_thread.c:188)
==8856==    by 0x560B6DA: start_thread (pthread_create.c:463)
==8856==    by 0x5944A3E: clone (clone.S:95)
==8856==  Address 0x2d0 is not stack'd, malloc'd or (recently) free'd
Number of cores 16, sandboxing cores 15 (start: 1) and module reqs 0
==8856== Thread 6:
==8856== Invalid read of size 4
==8856==    at 0x56132A6: pthread_kill (pthread_kill.c:40)
==8856==    by 0x409F66: software_interrupt_handle_signals (software_interrupt.c:72)
==8856==    by 0x561689F: ??? (in /lib/x86_64-linux-gnu/libpthread-2.27.so)
==8856==    by 0x56131F5: pthread_sigmask (pthread_sigmask.c:45)
==8856==    by 0x40A712: software_interrupt_unmask_signal (software_interrupt.h:86)
==8856==    by 0x40A712: worker_thread_main (worker_thread.c:188)
==8856==    by 0x560B6DA: start_thread (pthread_create.c:463)
==8856==    by 0x5944A3E: clone (clone.S:95)
==8856==  Address 0x2d0 is not stack'd, malloc'd or (recently) free'd
```

Which is basically this code:
```c
		/* A POSIX signal is delivered to one of the threads in our process.If sent by the kernel, "broadcast"
		 * by forwarding to all all threads */
		if (signal_info->si_code == SI_KERNEL) {
			for (int i = 0; i < runtime_total_worker_processors; i++) {
				if (pthread_self() != runtime_worker_threads[i]) {
					pthread_kill(runtime_worker_threads[i], SIGALRM);
				}
			}
		} else {
			/* If not sent by the kernel, this should be a signal forwarded from another thread */
			assert(signal_info->si_code == SI_TKILL);
		}
```

runtime_worker_threads is being indexed properly and seems to have sane pthread IDs
I have no clue about address 0x2d0

bushidocodes:
## Summary

Running `sledgert` under Valgrind/Memcheck produces a flood of false `Invalid read/write` errors — and with preemption on and multiple workers, hundreds of millions of them, including the report tracked in #69 (`software_interrupt_handle_signals` → `propagate_sigalrm` → `pthread_kill`). They are artifacts of three things Memcheck cannot model:

1. **Signal-based preemption** delivered while a worker runs on a switched sandbox stack.
2. **User-level context switching** (`swapcontext` / `arch_context`) — Valgrind prints `client switching stacks?` and loses track of stack memory.
3. **Direct `mmap` of sandbox linear memory** — large guard regions marked `noaccess` that the JIT'd wasm module then legitimately touches.

This PR adds tooling so the runtime can be checked under Valgrind without that noise. It does **not** change any runtime code.

## What's included

- **`runtime/tools/valgrind/sledge.supp`** — suppressions scoped to the false-positive classes only (signal machinery, JIT'd guest code, WASI shims touching guest memory, per-worker queues read after a context switch). It does not hide errors elsewhere in the control plane.
- **`runtime/tools/valgrind/run-valgrind.sh`** — wraps a `control-plane` mode (preemption off, one worker — what the existing per-test `make valgrind` targets use) and a `preemptive` mode, applying the suppressions and `LD_BIND_NOW` in both.
- **`docs/valgrind.md`** — explains why the false positives occur, how to run each mode, and the small documented residual.

```sh
runtime/tools/valgrind/run-valgrind.sh -d tests/multi-tenancy-sample                 # control-plane
runtime/tools/valgrind/run-valgrind.sh -m preemptive -w 8 -d tests/multi-tenancy-sample
```

## Why this is safe (no real bug is being hidden)

- The `pthread_t` passed to `pthread_kill` is a **valid mapped descriptor** — the worker-thread array is `calloc`'d, asserted non-zero before use, and the timer is armed only after every worker exists. (The original 2019 bug, a near-null/uninitialised slot, was already fixed.)
- `--track-origins=yes` shows **every** uninitialised-value report in the scheduler / run-queue / sandbox paths traces to a **single origin** — the hand-rolled stack setup in `arch_context_init` — with **zero heap-allocation origins**.
- A **differential leak run (20 vs 120 requests) is byte-for-byte identical**: no per-request leak. The few "lost" blocks are per-thread TLS/structures that aren't freed only because the runtime is stopped by a signal and never joins its threads.

With the suppressions applied, SLEdge's own runtime code is silent under Valgrind; the only remaining reports are unanchorable libc-level sandbox-memory and host-stdio noise that scales with guest execution.

Closes #69.

### DIFF
--- a/docs/valgrind.md
+++ b/docs/valgrind.md
@@ -1,0 +1,146 @@
+# Running SLEdge under Valgrind
+
+This document explains how to run the SLEdge runtime under Valgrind/Memcheck and
+why its output needs interpretation.
+
+## TL;DR
+
+- Valgrind reports many "Invalid read/write" errors against SLEdge that are
+  **false positives** — artifacts of how the runtime preempts, switches stacks,
+  and maps sandbox memory, none of which Memcheck can model.
+- For finding **real** memory bugs, use **control-plane mode** (preemption off,
+  one worker). This is what the per-test `make valgrind` targets already do.
+- To exercise the **preemptive** path (multiple workers + SIGALRM) under
+  Valgrind, use the suppression file that filters the known false positives.
+
+Both modes are wrapped by `runtime/tools/valgrind/run-valgrind.sh`.
+
+## Quick start
+
+```sh
+# Build the runtime + sample apps first (Docker route recommended; see README.md)
+
+# Control-plane mode: clean output, best for hunting real leaks/errors.
+runtime/tools/valgrind/run-valgrind.sh -d tests/multi-tenancy-sample
+
+# Preemptive mode: 8 workers + preemption, with false positives suppressed.
+runtime/tools/valgrind/run-valgrind.sh -m preemptive -w 8 -d tests/multi-tenancy-sample
+```
+
+Send requests from another shell while the runtime is up (e.g.
+`curl localhost:55555/admin?40`), then stop the runtime with Ctrl-C so Valgrind
+prints its summary.
+
+## Why Valgrind struggles with SLEdge
+
+Memcheck models a normal process. SLEdge violates three of its assumptions, and
+each produces a flood of bogus "Invalid read/write" reports:
+
+1. **Signal-based preemption.** A SIGALRM fires while a worker runs on a
+   sandbox's stack. The handler
+   (`software_interrupt_handle_signals` → `propagate_sigalrm` → `pthread_kill`)
+   reads the interrupted `ucontext` and the glibc thread descriptor (TCB).
+   Because the handler runs after a user-level stack switch, Valgrind has lost
+   track of those regions and flags every read.
+
+2. **User-level context switching** (`swapcontext` / `arch_context`). Valgrind
+   prints `client switching stacks? SP change ...` and can no longer tell valid
+   stack memory from garbage.
+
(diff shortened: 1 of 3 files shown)


emil916:
This is great! But I can't test this at this time. If you are sure this runs smoothly, then go ahead and merge.

PR #393, opened 2026-06-16
Size the work-stealing deque to its capacity and index it circularly

### LINKED ISSUE #89: Implement the Ability to Dynamically Resize Deque
The baseline design statically allocates a global request scheduler structure that can hold 524,288 sandbox requests. This choice was arbitrary. It also seems that the backing wrk data structure is always sized to DEQUE_MAX_SZ, so the size member used by the init function only acts as a soft max. This might be inefficient.

Relevant Code in `runtime/include/deque.h`
```
/* TODO: Implement the ability to dynamically resize! Issue #89 */
#define DEQUE_MAX_SZ (1 << 23)
```

and in `runtime/include/runtime.h`
```
#define RUNTIME_MAX_SANDBOX_REQUEST_COUNT (1 << 19) /* random! */
```

and in `global_request_scheduler_deque_initialize()` in `runtime/src/global_request_scheduler_deque.c`
```
	deque_init_sandbox(global_request_scheduler_deque, RUNTIME_MAX_SANDBOX_REQUEST_COUNT);
```

Note: if the dequeue is altered to be dynamically resizable, then it likely is worth investigating how a PQ-based global request queue could also be dynamically resized. The PQ interface have a `grow` API, but this has some design questions:

"Since the PQ has the grow API, we could in theory also resize the global request queue. However, since this is accessed by all cores, I am unclear how this would work. Can I protect a structure that might be realloced if the lock that protects it might get moved? I would assume this is unsafe. We could refactor our PQ from a VLA into a normal structure with a pointer to a buffer, but this might have performance implications. I suggest deferring this for a further effort since we likely would need to do some performance testing to understand the overhead (if any)." From https://github.com/gwsystems/sledge-serverless-framework/pull/325

bushidocodes:
## Summary

The work-stealing deque (`runtime/include/deque.h`, used by the FIFO global request scheduler) declared its backing store as a **fixed inline array** `type wrk[DEQUE_MAX_SZ]` — `1<<23` entries, i.e. **~64 MB** for `struct sandbox*` — while the `size` passed at init was only a *soft occupancy cap*. Two problems:

1. **Memory waste.** The buffer was always the compile-time maximum (`DEQUE_MAX_SZ`) regardless of the configured capacity (currently 4096). This is exactly the inefficiency #89 describes.
2. **Latent out-of-bounds.** `top`/`bottom` were used as raw, monotonically increasing indices with **no modulo wrap**. Since the global scheduler only `push`es (bottom++) and `steal`s (top++), the index marches forward forever and runs off the end of the array after `1<<23` total requests.

## Fix

Allocate the backing buffer on the heap at init, sized to the requested capacity (rounded up to a power of two), and index it **circularly** via a mask. The occupancy cap (`size - 1`) keeps the producer from lapping the consumers, so masked indices of live elements never collide. This makes memory scale with the configured capacity and removes the eventual overflow. The atomics and memory ordering are unchanged — only the index reads/writes gained `& mask`.

Also: add `deque_free`, have `deque_init` report allocation failure (now checked by the caller), and make `deque.h` self-contained (`stdbool`/`stdlib`/`errno`/`assert`).

**Not included (deliberately):** growing the buffer when full. Stealers consume the deque lock-free concurrently with the producer, so reallocating/freeing the backing buffer underneath an in-flight steal would be a use-after-free; a safe grow needs a lock-free resizable variant or synchronizing every stealer — the same concern raised for the PQ in #325. The deque returns `-ENOSPC` when full and the caller applies backpressure.

## Verification

- **Builds** clean.
- **FIFO end-to-end** (`SCHEDULER_FIFO`, multiple worker stealers + listener producer): resize requests return output **byte-identical** to the expected result (20/20).
- **Circular indexing unit test**: drives **1,000,007** push/steal pairs through a **size-8** ring — indices wrap ~125,000×, and the live window straddles the `7→0` boundary **750,000** times — all in strict FIFO order.
- **Memory**: `VmSize`/`VmData` drop **~64 MB** vs. the original (the deque now reserves the 4096-entry buffer instead of the `1<<23` array).

Closes #89.

### DIFF
--- a/runtime/include/deque.h
+++ b/runtime/include/deque.h
@@ -7,6 +7,12 @@
 #ifndef DEQUE_H
 #define DEQUE_H
 
+#include <assert.h>
+#include <errno.h>
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdlib.h>
+
 /*
  * This was implemented by referring to:
  * https://github.com/cpp-taskflow/cpp-taskflow/blob/9c28ccec910346a9937c40db7bdb542262053f9c/taskflow/executor/workstealing.hpp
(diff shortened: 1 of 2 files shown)


emil916:
LGTM!

Issue #224, opened 2021-05-18
sandboxes likely add themselves to the completion queue while running


bushidocodes:
While going through the latest PR, it appeared that sandboxes are adding themselves to the completion queue while running by calling `sandbox_exit`. This requires some extra study. I have a strong suspicion that this might be the cause of the segfault as described in #66, so if this is changed, the hack should also be removed and the code should be tested for stability.

It is worth assessing if the distinction between the RETURNED and COMPLETED state is actually necessary. There would be value in simplification here.

bushidocodes:
This is not currently a bug because preemption is only enabled in the `RUNNING_USER` sandbox state. This means that the completion queue will not be cleaned up until the next preemption or yield when a sandbox is switched away from.

Issue #219, opened 2021-05-07
global_request_scheduler_deque_mutex is never used


bushidocodes:
Is this a mistake or can it be removed?

bushidocodes:
Duplicate of #291

Issue #66, opened 2020-07-09
Segmentation Fault on sandbox exit after preemption


bushidocodes:
The runtime currently experiences a mysterious segfault when a sandbox is preempted for another with a earlier deadline while running the minheap-based request queues and runqueues..

This remains unsolved, but a temporary band-aide has been added which prevents the segfault. See the section `Temporary Workaround` below for details. Be sure to comment out `assert(0)` if you want to recreate this bug.

This can be reproduced by the test infrastructure in `runtime/tests/preemption.` Start `debug.sh` to start the runtime under GDB, and then run `client1.sh`. This will run indefinitely and should not encounter an error. Then run `client2.sh`, which adds requests with an earlier deadline, forcing preemption.

In GDB, this triggers an error roughly as follows:

```
Thread 4 "awsmrt" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7ffff5bd1700 (LWP 22007)]
0x00000000004092c1 in current_sandbox_main () at src/sandbox.c:278
278             worker_thread_on_sandbox_exit(sandbox);
(gdb) back
#0  0x00000000004092c1 in current_sandbox_main () at src/sandbox.c:278
#1  0x0000000000000000 in ?? ()
(gdb)
```

Where `sandbox.c:278` is as follows:
```c
worker_thread_on_sandbox_exit(sandbox);
```

Based on an investigation of `objdump -Shrtl`, this is compiled into the following assembly:

```
current_sandbox_main():
/awsm/runtime/src/sandbox.c:278
  4092b5:       4c 89 e7                mov    %r12,%rdi
  4092b8:       48 83 c4 48             add    $0x48,%rsp
  4092bc:       5b                      pop    %rbx
  4092bd:       41 5c                   pop    %r12
  4092bf:       41 5d                   pop    %r13
  4092c1:       41 5e                   pop    %r14
  4092c3:       41 5f                   pop    %r15
  4092c5:       5d                      pop    %rbp
  4092c6:       e9 05 18 00 00          jmpq   40aad0 <worker_thread_on_sandbox_exit>
  4092cb:       bf 31 78 41 00          mov    $0x417831,%edi
  4092d0:       be 95 77 41 00          mov    $0x417795,%esi
  4092d5:       b9 dd 78 41 00          mov    $0x4178dd,%ecx
  4092da:       ba 3a 00 00 00          mov    $0x3a,%edx
[...]

PR #395, opened 2026-06-17
Deepen TCP listen backlog to cut connection-setup tail latency

### LINKED ISSUE #258: Poor Tail Latency in Concurrency Experiment
The `concurrency` experiment executes the "empty" workload (which calls `printf("S\n")` and immediately exits) with high levels of concurrency ranging from 1 to 100. The pattern of how hey is invoked here exhibits extreme bursts of requests. 0 -> 1 -> 0 -> 20 -> 0 -> 40 -> 0 ->  60 -> 0 -> 80 -> 0 -> 100. The spec.json defines a single module, so all requests have the same relative deadline. As such, FIFO and EDF should behave similarly.

During a recent run of this experiment, I witnessed p100 tail latency across all scheduling policies of up to 20s. The logs and charts for this runs are as follows:
 
[server_concurrency_gnuplots.zip](https://github.com/gwsystems/sledge-serverless-framework/files/6649334/server_concurrency_gnuplots.zip)

Note that based on the "offset" reported by hey, it appears that the worst tail latency occurs in batches of roughly 10ms.

Based on all this, I suspect that (assuming this is reproducible externally), there is a context/sandbox switching bug. I believe that we aren't seeing it in other experiments because this experiment triggers a significantly higher rate of sandbox switches due to:
* "empty" workload executing to completion very quickly
* The bursty request pattern of the driver script.

My debugging suggestion for this bug would be to validate the bug as follows:
1. Sanity Check hey parameters
2. Run on CloudLab to see if this relates to my hacky home office setup
3. Refactor experiment to use server-side reporting
4. If this is visible with server-side metrics on cloudlab, then there is likely a bug
5. I suspect that focusing on the state of the system during the 10ms of reported long running tasks is a good bet. The question is why these requests were unable to propagate through the system? Did a worker spin? Did the sandboxes block and get caught in epoll? Looking at the time of sandbox states for these sandboxes might help.

Possibly relevant issues:
- #224
- #219 
- #66

bushidocodes:
## Summary

A single listener thread accepts **every** client connection. Under a burst of short-lived, non-keep-alive requests — exactly what the `concurrency` experiment generates driving the `empty` workload — the 128-entry accept queue overflows, the kernel drops the connection, and the client only retries after the **~1s initial SYN RTO**, producing a sharp ~1s tail-latency spike.

This raises the listen backlog from 128 to 4096. The kernel silently caps it to `/proc/sys/net/core/somaxconn`, so requesting more than the host allows is harmless. The now-redundant compile-time truncation `#warning` is removed and the rationale folded into the comment. Recent Linux raised the default somaxconn to 4096 (kernel 5.4+).

## Verification

`hey` against the `empty` workload, c=200, 50k requests:

| | backlog 128 | backlog 4096 |
|---|---|---|
| p99.9 | 1084 ms | **283 ms** |
| max | 1109 ms | **292 ms** |
| requests >1s | **202** | **0** |

At concurrency 1–100 (the experiment's actual range) the tail was already clean in both cases — this removes the spike that appears once the connection-setup burst exceeds the old backlog.

## Context (#258)

While investigating #258 ("Poor Tail Latency in Concurrency Experiment"), the reported ~20s p100 tail **did not reproduce** on current `master` — across concurrency 1→100, EDF and FIFO, 120k requests, the worst case was ~101ms with zero requests over 1s. The switching/lifecycle bugs that issue suspected (#224, #66) and the unused deque mutex (#219) are already fixed, and the request path was refactored since. The one real, reproducible tail contributor that remained was this connection-setup backlog, fixed here.

A deeper limit — the single listener thread's accept/parse/send throughput — remains as potential future work; this change absorbs connection-setup bursts but does not raise steady-state accept throughput.

Closes #258.

### DIFF
--- a/runtime/include/tcp_server.h
+++ b/runtime/include/tcp_server.h
@@ -13,19 +13,17 @@
 #include "likely.h"
 
 /*
- * Defines the listen backlog, the queue length for completely established socketeds waiting to be accepted
- * If this value is greater than the value in /proc/sys/net/core/somaxconn (typically 128), then it is silently
- * truncated to this value. See man listen(2) for info
+ * Defines the listen backlog: the queue length for completely established sockets waiting to be accepted.
+ * The kernel silently caps this to /proc/sys/net/core/somaxconn, so requesting more than the host allows is
+ * harmless (it just truncates). We request a large value because a single listener thread accepts every
+ * connection; under a burst of short-lived, non-keep-alive requests a small backlog overflows, the kernel
+ * drops the connection, and the client only retries after the ~1s initial SYN RTO — a sharp ~1s tail-latency
+ * spike. A deeper backlog absorbs the burst and removes that spike.
  *
- * When configuring the number of sockets to handle, the queue length of incomplete sockets defined in
- * /proc/sys/net/ipv4/tcp_max_syn_backlog should also be considered. Optionally, enabling syncookies removes this
- * maximum logical length. See tcp(7) for more info.
+ * The queue length of incomplete (half-open) sockets in /proc/sys/net/ipv4/tcp_max_syn_backlog should also be
+ * considered; enabling syncookies removes that maximum. See listen(2) and tcp(7).
  */
-#define TCP_SERVER_MAX_PENDING_CLIENT_REQUESTS 128
-#if TCP_SERVER_MAX_PENDING_CLIENT_REQUESTS > 128
-#warning \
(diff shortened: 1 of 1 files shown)


bushidocodes:
My dev setup in virtualized and noisy, so don't trust my numbers. I suspect there should be some improvements on a native environment though. I assume you may need to increase some of the parameters to see this tail on native Linux.

emil916:
Holy cow! I didn't know about the existence of such a "knob" in the system. It has likely had impacts on all sledge experiments to some degree.

Issue #290, opened 2021-12-10
Consider setting a dedicated signal handler stack to make sandbox stack overflows recoverable


bushidocodes:
```c
runtime/include/wasm_stack.h
	int rc        = munmap(stack->buffer, stack->capacity + PAGE_SIZE);
	stack->buffer = NULL;
	if (unlikely(rc == -1)) perror("munmap");
}
 ```
Member
@gparmer gparmer yesterday
You have any idea what happens if we set up a signal for SIG_TRAP, then trap out by hitting the guard page here? Normally the signal would execute on the stack....but....

 
Member
Author
@bushidocodes bushidocodes 21 hours ago
@gparmer - I would assume the signal handler would panic the process or something. I seem to recall that you can consider signal handlers to use a dedicated stack. Perhaps we should be doing that so a module stack overflow is recoverable?

bushidocodes:
I thought through this a bit more. Currently, large objects on the stack are most likely going to end up on the stack placed by WASI-SDK within a sandbox's linear memory. If a stack overflow occurs there, the stack corrupts the data segment unless `wasm-ld` was configured to place the stack at the front of linear memory, in which case, the stack pointer would be decremented below 0 and overflow to some high address that would likely be beyond what the size of the linear memory, triggering a WebAssembly trap. 

In contrast, the values on this stack are the WebAssembly locals, parameters, and operand stack generated by `WASI-SDK` or whatever other compiler generated the module. Given that this code is highly optimized, I assume that this pressure is less critical. `sledgert` has some parameters to customize this stack size, but we hardcode this to 512KB, which I consider a substantial overallocation that ameliorates the risk of the stack overflow situation discussed above.

If WebAssembly ends up adopting an object model that ends up shifting state from the in-memory stack to WebAssembly locals, this might change this dynamic. Otherwise, I don't think overflows are that big of a concern compared to the stack inside linear memory.

If we do want to deal with this, `signaltstack(2)` essentially has the complete sample code for setting up an alternate signal stack to deal with SIGSEGV.

PR #398, opened 2026-06-17
[Experimental] Alternate signal stack to make sandbox stack overflows recoverable

bushidocodes:
> **Draft / Experimental.** The mechanism is validated, but a real in-runtime sandbox overflow has not been exercised end to end (see Verification). Opening for discussion before promoting.

## Problem

Each sandbox runs on its own `mmap`'d native stack with a `PROT_NONE` guard page below it (`wasm_stack.h`). The signal handlers are registered with only `SA_SIGINFO | SA_RESTART` — **no `sigaltstack`**. So when a sandbox overflows its stack and faults on the guard page, there's no room on the exhausted stack to deliver SIGSEGV, and the kernel kills the whole runtime instead of trapping just that sandbox. The recovery machinery already exists (`current_sandbox_start` does `sigsetjmp`; the handler does `current_sandbox_trap` → `siglongjmp` back to a valid part of the stack) — it just can't run because the signal can't be delivered. (Addresses #290.)

## Change

- Each worker registers a per-thread **alternate signal stack** (`sigaltstack`) in `worker_thread_main`, before unmasking the fault signals.
- The synchronous fault signals **SIGSEGV and SIGFPE** are registered `SA_ONSTACK`, so they're delivered on the alt stack even when the sandbox stack is exhausted → the existing `siglongjmp` recovery runs.
- **SIGALRM / SIGUSR1 are intentionally left off the alt stack.** They context-switch away (non-local exit) rather than returning, which doesn't compose with an alternate stack; isolating `SA_ONSTACK` to the fault handlers keeps the preemption path untouched.

3 files, +44 lines.

## Verification

- **Mechanism proof** — a standalone test reproducing the runtime's exact primitives (guard-paged `mmap` stack run via `makecontext`/`swapcontext`, `sigsetjmp` recovery, handler `siglongjmp`): **without** the alt stack the process is killed by SIGSEGV (exit 139); **with** it, the overflow is `RECOVERED` (exit 0).
- **No regression** — normal workload (empty, EDF + preemption, 20,000 requests) → 20,000/20,000 `200`s, runtime alive, no crashes; confirms the alt stack + selective `SA_ONSTACK` doesn't disturb preemption.
- Builds clean.

### Why experimental / not fully verified

A real in-`sledgert` overflow was **not** exercised end to end: no prebuilt module recurses deeply enough to exhaust the 512 KB native stack (per #290 this is hard to trigger — wasm locals live in linear memory, so it needs deep recursion), and building a custom deeply-recursive module goes through the libsledge `.wasm.so` toolchain. The change is sound by composition (the recovery path pre-exists and already handles traps like linear-memory-OOB SIGSEGV → 500; this only adds alt-stack *delivery*, proven by the mechanism test), but I'd want a real overflow test before marking it non-experimental.

Relates to #290.

### DIFF
--- a/runtime/include/software_interrupt.h
+++ b/runtime/include/software_interrupt.h
@@ -92,6 +92,11 @@ software_interrupt_deferred_sigalrm_replay()
  * Exports from software_interrupt.c *
  ************************/
 
+/* Size of each worker's alternate signal stack. The SA_ONSTACK handlers (SIGSEGV, SIGFPE) only siglongjmp out,
+ * so this needs little space; keep it comfortably above MINSIGSTKSZ. */
+#define SOFTWARE_INTERRUPT_ALT_STACK_SIZE (64 * 1024)
+
+void software_interrupt_alt_stack_initialize(void);
 void software_interrupt_arm_timer(void);
 void software_interrupt_cleanup(void);
 void software_interrupt_disarm_timer(void);
--- a/runtime/src/software_interrupt.c
+++ b/runtime/src/software_interrupt.c
@@ -278,6 +278,19 @@ software_interrupt_initialize(void)
 		/* Mask this signal on the listener thread */
 		software_interrupt_mask_signal(signal);
 
+		/* Deliver the synchronous fault signals (SIGSEGV, SIGFPE) on the per-worker alternate signal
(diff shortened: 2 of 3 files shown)


emil916:
This is interesting! We can definitely circle back to this later. So, yes, agreed to keep it open for now.

PR #402, opened 2026-06-23
Document that modules must be WASI modules (#329)

### LINKED ISSUE #329: Document and validate that aWSM generated *.bc is a WASI module
aWsm has added support for modules outside of the WASI spec. For example, modules that do not use the linear memory. However, SLEdge seems to be more opinionated around WASI, and it is unclear to me if supporting non-WASI WebAssembly targets has any particular value.

Our documentation should clarify our assumptions around WASI symbols.

bushidocodes:
Closes #329.

## Background

aWsm can emit WebAssembly outside the WASI spec (e.g. modules that don't use linear memory), but SLEdge is opinionated around WASI. That assumption was undocumented.

The "validate during dlopen" half of the issue is **already satisfied** by existing code: `sledge_abi_symbols_init` resolves the aWsm-generated symbols and fails the load when the mandatory linear-memory symbols (`sledge_abi__init_mem`, `sledge_abi__wasm_memory_starting_pages`/`_max_pages`) are absent, which is exactly the non-WASI / no-linear-memory case. So this PR is docs-only.

## Changes

**`docs/WASI.md`** — new "Module Requirements and Assumptions" section:
- Modules must be WASI modules compiled by aWsm and linked against `libsledge`; non-WASI aWsm output is unsupported.
- **Linear memory is mandatory** — SLEdge's stdin/stdout I/O model lives in linear memory, so a module without it is rejected at load time when its aWsm memory symbols don't resolve.
- Explains why `wasi_snapshot_preview1_*` symbol presence is *not* a usable WASI signal: those shims come from `libsledge` and are unconditionally exported from every `.so` via `--whole-archive --export-dynamic`.
- Records the unresolved `--runtime-globals` question (previously only an inline code comment).

No runtime code changes.

### DIFF
--- a/docs/WASI.md
+++ b/docs/WASI.md
@@ -1,5 +1,20 @@
 SLEdge only implemented a subset of the WASI syscall interface 
 
+## Module Requirements and Assumptions
+
+SLEdge assumes that every module it loads is a **WASI module** compiled by aWsm and linked against `libsledge`. aWsm can also emit WebAssembly outside the WASI spec (for example, modules that do not use linear memory), but such modules are **not** supported by SLEdge.
+
+Concretely, a module must satisfy the following to run:
+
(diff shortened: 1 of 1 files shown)


bushidocodes:
runtime/include/sledge_abi_symbols.h already had the checks this issue discussed, so this is a doc-only change.

